PR #1, opened 2024-09-21
ci: introduce github actions

chr-hertel:
not yet a matrix, but a start

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -1,0 +1,36 @@
+name: pipeline
+on: pull_request
+
+jobs:
+  pipeline:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,36 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.2'
+
+      - name: Cache Vendors
```
OskarStark, line 16: Instead of using this, just use Ramsey/composer-install action

chr-hertel: thanks for that pointer :) see #4

PR #5, opened 2024-09-21
feat: implement universal chain

chr-hertel:
idea is to bring all features into one universal chain and remove the other ones.


src/Anthropic/Model/Claude.php
```diff
@@ -19,16 +21,28 @@ public function __construct(
...
 
-    public function call(MessageBag $messages, array $options = []): array
+    public function call(MessageBag $messages, array $options = []): Response
     {
         $system = $messages->getSystemMessage();
 
-        return $this->runtime->request([
+        $response = $this->runtime->request([
             'model' => $this->version->value,
             'temperature' => $this->temperature,
```
OskarStark, line 30: topP should also be configurable

chr-hertel: oh, `$options` is ignored completely at the moment, yes

PR #8, opened 2024-09-21
feat: adding youtube transcription tool

chr-hertel:


README.md
```diff
@@ -105,6 +106,18 @@ Depending on the example you need to export needed environment variables for API
    php examples/toolchain-serpapi.php
    ```
 
+1. Weather Tool
+   ```bash
+   export OPENAI_API_KEY=sk-...
+   php examples/toolchain-weather.php
```
OskarStark, line 112: shouldn't we talk consistently about `toolbox` instead of `toolchain`?

composer.json
```diff
@@ -29,12 +29,16 @@
...
+        "symfony/dom-crawler": "^6.4 || ^7.1",
         "symfony/var-dumper": "^6.4 || ^7.1"
     },
     "suggest": {
         "codewithkyrian/chromadb-php": "For using the ChromaDB as retrieval vector store.",
         "probots-io/pinecone-php": "For using the Pinecone as retrieval vector store.",
-        "symfony/clock": "For using the clock tool."
+        "symfony/clock": "For using the clock tool.",
+        "symfony/css-selector": "For using the YouTube transcription tool.",
+        "symfony/dom-crawler": "For using the YouTube transcription tool."
```
OskarStark, line 41: In Symfony we avoid/banned using suggestions, instead we decide on runtime, if we need something and I like it more

README.md
chr-hertel: Yea, good point, already legacy after that universal chain refactoring

PR #10, opened 2024-09-21
chore: tests for simple value classes

chr-hertel:


tests/Message/MessageTest.php
```diff
@@ -0,0 +1,136 @@
...
+use PHPUnit\Framework\Attributes\Small;
+use PHPUnit\Framework\Attributes\UsesClass;
+use PHPUnit\Framework\TestCase;
+
+#[CoversClass(Message::class)]
+#[UsesClass(ToolCall::class)]
+#[Small]
+final class MessageTest extends TestCase
+{
+    public function testCreateSystemMessage(): void
```
OskarStark, line 20: We could use Test Attribute instead of the prefix in the method name. Interested? If yes I can provide a PR

tests/Message/MessageBagTest.php
```diff
@@ -0,0 +1,108 @@
...
+{
+    public function testGetSystemMessage(): void
+    {
+        $messageBag = new MessageBag(
+            Message::forSystem('My amazing system prompt.'),
+            Message::ofAssistant('It is time to sleep.'),
+            Message::ofUser('Hello, world!'),
+        );
+
+        $systemMessage = $messageBag->getSystemMessage();
```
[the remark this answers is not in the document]
chr-hertel: Not sure it should. I'd rather say the MessageBag should make sure there is only one and that it's the first.

tests/Message/MessageTest.php
chr-hertel: Interesting, never ever used that. What's pro or con? Or rather style?

tests/Message/MessageBagTest.php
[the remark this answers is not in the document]
OskarStark: When answering to a function call I create 2 messages:

1) -> incoming function call
2) <- system: please show only name, price and the product url from the following product data
3) <- json encoded string of a product

-> call() the llm

so I decided in my world, I always return a MessageBag from my function responses and append them.

I also compose my initial chat with many SYSTEM messages:
1) The time is now.....
2) You are a helpful assistant
3) etc.

tests/Message/MessageTest.php
OskarStark: Rather style, as I switched from `@test` annotation to `#[Test]` attribute

PR #12, opened 2024-09-22
Introduce `Version` value object

OskarStark:


composer.json
```diff
@@ -19,7 +19,8 @@
         "symfony/property-info": "^6.4 || ^7.1",
         "symfony/serializer": "^6.4 || ^7.1",
         "symfony/type-info": "^6.4 || ^7.1",
-        "symfony/uid": "^6.4 || ^7.1"
+        "symfony/uid": "^6.4 || ^7.1",
+        "webmozart/assert": "^1.11"
```
OskarStark, line 23: Not needed, but I usually use this lib to assert things in my value objects, we can ofc go with plain php if you want to avoid the dep

src/Anthropic/Model/Claude.php
```diff
@@ -15,17 +15,18 @@ final class Claude implements LanguageModel
 {
     public function __construct(
         private ClaudeRuntime $runtime,
-        private Version $version = Version::SONNET_35,
+        private ?Model $model = null,
```
OskarStark, line 18: Not sure about the name `Model`, because now we have `Model` and `LanguageModel` 🤔 

Or we keep `Version` and remove the enums.

src/Anthropic/Model/Model.php
```diff
@@ -0,0 +1,27 @@
...
+    ) {
+        Assert::stringNotEmpty($name);
+    }
+
+    public function getName(): string
+    {
+        return $this->name;
+    }
+
+    public static function fromVersion(Version $version): self
```
OskarStark, line 23: This way we can easily maintain some things, for example that 2 models support structured output

src/OpenAI/Model/Model.php
```diff
@@ -0,0 +1,41 @@
...
+    {
+        return $this->name;
+    }
+
+    public function supportsStructuredOutput(): bool
+    {
+        return $this->supportStructuredOutput;
+    }
+
+    public static function fromVersion(GptVersion|EmbeddingsVersion $version): self
```
OskarStark, line 33: Or we crate one Model class for embeddings and one for GPT

src/Anthropic/Model/Claude.php
OskarStark: Maybe `ModelVersion`?

composer.json
chr-hertel: yea, mixed but no hard feelings :D

src/Anthropic/Model/Claude.php
chr-hertel: yes, how about `ModelVersion` and inlining the enums as constants that are used for named constructor?

like `ModelVersion::sonnet35()` or `ModelVersion::gpt4o()`

WDYT?

OskarStark: Will rework this PR to get a feeling

composer.lock
chr-hertel, line 1: you reverting your own PR here :see_no_evil:

but happened to me as well, but that why i pushed #22

OskarStark: Indeed :-D

OskarStark: Fixed

src/OpenAI/Model/Gpt/Version.php
```diff
@@ -4,19 +4,57 @@
...
-
-    public function supportsStructuredOutput(): bool
-    {
-        return self::GPT_4o === $this || self::GPT_4o_MINI === $this;
+    /**
+     * @param non-empty-string $name
+     */
+    public function __construct(
+        public string $name,
+        public bool $supportStructuredOutput,
```
chr-hertel, line 16: can we add `readonly` or do i miss sth?

```suggestion
        public readonly string $name,
        public readonly bool $supportStructuredOutput,
```

chr-hertel: basically same questions for the other ones.

i guess on class level it's not possible due to named constructors, right?

chr-hertel:
> WDYT @chr-hertel ?

yes, looks pretty good to me :)

src/OpenAI/Model/Gpt/Version.php
OskarStark: done

chr-hertel (approved):
Nice one, thanks! :pray:

PR #13, opened 2024-09-22
Rename method

OskarStark:
I think it is more readable, feel free to merge or close


chr-hertel:
yup, good one - extended it to the interface as well :+1:

chr-hertel (approved):

PR #16, opened 2024-09-22
Simplify constructing `ToolAnalyzer`

OskarStark:
Not sure about this PR, maybe its better to be explicit?

WDYT? Feel free to close


chr-hertel (approved):
I think it's fine since that used service is quite simple, has no interface and is still part of the constructor arguments :+1:

PR #17, opened 2024-09-22
Helpful exceptions

OskarStark:


composer.json
```diff
@@ -36,9 +36,7 @@
     "suggest": {
         "codewithkyrian/chromadb-php": "For using the ChromaDB as retrieval vector store.",
         "probots-io/pinecone-php": "For using the Pinecone as retrieval vector store.",
-        "symfony/clock": "For using the clock tool.",
-        "symfony/css-selector": "For using the YouTube transcription tool.",
-        "symfony/dom-crawler": "For using the YouTube transcription tool."
```
chr-hertel, line 41: yea i know symfony does that, but the use case there is different i feel.

a) symfony usually comes with flex and has those interdependencies all over the place
b) with notifier and mailer symfony avoid those situations with having separate vendor specific adapter

src/ToolBox/Tool/YouTubeTranscriber.php
chr-hertel, line 1: :+1:

composer.json
chr-hertel: or is there something else i miss here? :thinking:

OskarStark: Flex does not help in that case, how should it know if it should install css-selector for example?

b) no they don't, as a bridge is just a package in the end, which can behave differently. a bridge could for example use Curl or another http client internally

chr-hertel: No, i don't meant that Flex actually helps - but the way Symfony nowadays is sliced and flexible from a "which package is there?" point of view, is due to Flex and that restructuring with SF4. Would totally be overkill to have `suggest` all over the place with the Symfony components relying on each other based on specific use cases - that's why that runtime based approach a la "this execution path is only possible with `composer req symfony/...`" is way better in symfony's use case.

decoupling into a bridge enables symfony notifier not to suggest anything, but the concrete adapter can just install.

i def like the additional context in the exception, but don't see - at least with this style of package design - no benefit in skipping the suggestions

OskarStark: Got it

OskarStark:
Ready to merge

chr-hertel (approved):

PR #18, opened 2024-09-22
Add `rector/rector`

OskarStark:


rector.php
```diff
@@ -0,0 +1,24 @@
...
+        __DIR__.'/tests',
+    ])
+    ->withPhpSets(php82: true)
+    ->withSets([
+        PHPUnitSetList::PHPUNIT_110,
+        PHPUnitSetList::ANNOTATIONS_TO_ATTRIBUTES,
+    ])
+    ->withImportNames(importNames: true, importShortClasses: false)
+    ->withSkip([
+        ClosureToArrowFunctionRector::class,
```
OskarStark, line 22: I don't like arrow functions 😄

chr-hertel (approved):
will need to get used to it, but def safe enough to try :)

PR #24, opened 2024-09-23
More specific exception

OskarStark:


src/Exception/MessageBagNotFound.php
```diff
@@ -6,7 +6,7 @@
 
 final class MessageBagNotFound extends \DomainException
```
OskarStark, line 7: Is this class used somewhere?

chr-hertel: no, good catch, the caller is gone since a while.

it was an early MessageStore implementation:
https://github.com/php-llm/llm-chain/blob/4804cf0fc75130a8eb76e03c77cb4a6c6a4cb4cc/src/Message/MessageStore.php

chr-hertel (approved):

PR #25, opened 2024-09-23
Message Splitting to allow multiple message implementations without stacking arguments

DZunke:
Hey,

i utilize the library to also generate image descriptions by a delivered picture. With the MessageBag this is currently not possible because only textual messages are supported but for image describing one needs to hand over the image to gpt. As with the nice improvements from the weekend for the tooling information i saw that the message stack up with more and more arguments. 

So i thought about splitting up the messages into more speaking single objects with an purpose for each. What do you think about the idea to split it more up? As an example there is an image description generator script now.

This MR is a absolute basic. That is why i have kept the Message object as a bridge at first for the specific messages. But with testing, if the idea is fine, i would remove it and rewrite all the code that depends on it.

### DIFF
--- a/examples/image-describer.php
+++ b/examples/image-describer.php
@@ -1,13 +1,13 @@
 <?php
 
 use PhpLlm\LlmChain\Chain;
+use PhpLlm\LlmChain\Message\Content\Image;
 use PhpLlm\LlmChain\Message\Message;
 use PhpLlm\LlmChain\Message\MessageBag;
 use PhpLlm\LlmChain\OpenAI\Model\Gpt;
 use PhpLlm\LlmChain\OpenAI\Model\Gpt\Version;
 use PhpLlm\LlmChain\OpenAI\Runtime\OpenAI;
 use Symfony\Component\HttpClient\HttpClient;
-use PhpLlm\LlmChain\Message\Content\Image;
 
 require_once dirname(__DIR__).'/vendor/autoload.php';
 
@@ -20,7 +20,7 @@ $messages = new MessageBag(
     Message::ofUser(
         'Describe the images as a comedian would do it.',
         new Image('https://upload.wikimedia.org/wikipedia/commons/thumb/3/31/Webysther_20160423_-_Elephpant.svg/350px-Webysther_20160423_-_Elephpant.svg.png'),
-		new Image('https://upload.wikimedia.org/wikipedia/commons/thumb/3/37/African_Bush_Elephant.jpg/320px-African_Bush_Elephant.jpg'),
+        new Image('https://upload.wikimedia.org/wikipedia/commons/thumb/3/37/African_Bush_Elephant.jpg/320px-African_Bush_Elephant.jpg'),
     ),
 );
 $response = $chain->call($messages);
--- a/examples/image-describer.php
+++ b/examples/image-describer.php
@@ -1,0 +1,27 @@
+<?php
+
+use PhpLlm\LlmChain\Chain;
+use PhpLlm\LlmChain\Message\Message;
+use PhpLlm\LlmChain\Message\MessageBag;
+use PhpLlm\LlmChain\OpenAI\Model\Gpt;
+use PhpLlm\LlmChain\OpenAI\Model\Gpt\Version;
+use PhpLlm\LlmChain\OpenAI\Runtime\OpenAI;
+use Symfony\Component\HttpClient\HttpClient;
+
+require_once dirname(__DIR__).'/vendor/autoload.php';
+
+$runtime = new OpenAI(HttpClient::create(), getenv('OPENAI_API_KEY'));
+$llm = new Gpt($runtime, Version::gpt4oMini());
+
+$chain = new Chain($llm);
+$messages = new MessageBag(
+    Message::forSystem('You are an image analyzer that looks to images like a comedian would like.'),
+    Message::ofUser(
+        'Describe the image as a comedian would do it.',
+        'https://upload.wikimedia.org/wikipedia/commons/thumb/3/31/Webysther_20160423_-_Elephpant.svg/350px-Webysther_20160423_-_Elephpant.svg.png',
+        'https://upload.wikimedia.org/wikipedia/commons/thumb/3/37/African_Bush_Elephant.jpg/320px-African_Bush_Elephant.jpg',
+    ),
+);
+$response = $chain->call($messages);
+
+echo $response.PHP_EOL;
--- a/examples/image-describer.php
+++ b/examples/image-describer.php
@@ -15,9 +15,9 @@ $llm = new Gpt($runtime, Version::gpt4oMini());
 
 $chain = new Chain($llm);
 $messages = new MessageBag(
-    Message::forSystem('You are an image analyzer that looks to images like a comedian would like.'),
+    Message::forSystem('You are an image analyzer bot that helps identify the content of images.'),
     Message::ofUser(
-        'Describe the image as a comedian would do it.',
+        'Describe the images as a comedian would do it.',
         'https://upload.wikimedia.org/wikipedia/commons/thumb/3/31/Webysther_20160423_-_Elephpant.svg/350px-Webysther_20160423_-_Elephpant.svg.png',
         'https://upload.wikimedia.org/wikipedia/commons/thumb/3/37/African_Bush_Elephant.jpg/320px-African_Bush_Elephant.jpg',
     ),
--- a/examples/image-describer.php
+++ b/examples/image-describer.php
@@ -7,6 +7,7 @@ use PhpLlm\LlmChain\OpenAI\Model\Gpt;
 use PhpLlm\LlmChain\OpenAI\Model\Gpt\Version;
 use PhpLlm\LlmChain\OpenAI\Runtime\OpenAI;
 use Symfony\Component\HttpClient\HttpClient;
+use PhpLlm\LlmChain\Message\Content\Image;
 
 require_once dirname(__DIR__).'/vendor/autoload.php';
 
@@ -18,8 +19,8 @@ $messages = new MessageBag(
     Message::forSystem('You are an image analyzer bot that helps identify the content of images.'),
     Message::ofUser(
         'Describe the images as a comedian would do it.',
-        'https://upload.wikimedia.org/wikipedia/commons/thumb/3/31/Webysther_20160423_-_Elephpant.svg/350px-Webysther_20160423_-_Elephpant.svg.png',
-        'https://upload.wikimedia.org/wikipedia/commons/thumb/3/37/African_Bush_Elephant.jpg/320px-African_Bush_Elephant.jpg',
+        new Image('https://upload.wikimedia.org/wikipedia/commons/thumb/3/31/Webysther_20160423_-_Elephpant.svg/350px-Webysther_20160423_-_Elephpant.svg.png'),
+		new Image('https://upload.wikimedia.org/wikipedia/commons/thumb/3/37/African_Bush_Elephant.jpg/320px-African_Bush_Elephant.jpg'),
     ),
 );
 $response = $chain->call($messages);
--- a/src/Message/Content/Image.php
+++ b/src/Message/Content/Image.php
@@ -4,9 +4,9 @@ declare(strict_types=1);
 
 namespace PhpLlm\LlmChain\Message\Content;
 
-final readonly class ImageUrlContent implements ContentInterface
+final readonly class Image implements ContentInterface
 {
-    public function __construct(public string $imageUrl)
+    public function __construct(public string $image)
     {
     }
 
@@ -15,6 +15,6 @@ final readonly class ImageUrlContent implements ContentInterface
      */
     public function jsonSerialize(): array
     {
-        return ['type' => 'image_url', 'image_url' => ['url' => $this->imageUrl]];
+        return ['type' => 'image_url', 'image_url' => ['url' => $this->image]];
     }
 }
--- a/src/Message/Content/Image.php
+++ b/src/Message/Content/Image.php
@@ -6,7 +6,10 @@ namespace PhpLlm\LlmChain\Message\Content;
 
 final readonly class Image implements ContentInterface
 {
-    public function __construct(public string $image)
+    /**
+     * @param string $url An URL like "http://localhost:3000/my-image.png" or a data url like "data:image/png;base64,iVBOR[...]"
+     */
+    public function __construct(public string $url)
     {
     }
 
@@ -15,6 +18,6 @@ final readonly class Image implements ContentInterface
      */
     public function jsonSerialize(): array
     {
-        return ['type' => 'image_url', 'image_url' => ['url' => $this->image]];
+        return ['type' => 'image_url', 'image_url' => ['url' => $this->url]];
     }
 }
--- a/src/Message/Content/ImageUrlContent.php
+++ b/src/Message/Content/ImageUrlContent.php
@@ -1,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpLlm\LlmChain\Message\Content;
+
+final readonly class ImageUrlContent implements ContentInterface
+{
+    public function __construct(public string $imageUrl)
+    {
+    }
+
+    /**
+     * @return array{type: 'image_url', image_url: array{url: string}}
+     */
+    public function jsonSerialize(): array
+    {
+        return ['type' => 'image_url', 'image_url' => ['url' => $this->imageUrl]];
+    }
+}
--- a/src/Message/Message.php
+++ b/src/Message/Message.php
@@ -4,94 +4,37 @@ declare(strict_types=1);
 
 namespace PhpLlm\LlmChain\Message;
 
+use PhpLlm\LlmChain\Message\Content\ImageUrlContent;
+use PhpLlm\LlmChain\Message\Content\TextContent;
 use PhpLlm\LlmChain\Response\ToolCall;
 
-final readonly class Message implements \JsonSerializable
+final readonly class Message
 {
-    /**
-     * @param ?ToolCall[] $toolCalls
-     */
-    public function __construct(
-        public ?string $content,
-        public Role $role,
-        public ?array $toolCalls = null,
-    ) {
+    // Disabled by default, just a bridge to the specific messages
+    private function __construct()
+    {
     }
 
-    public static function forSystem(string $content): self
+    public static function forSystem(string $content): SystemMessage
     {
-        return new self($content, Role::System);
+        return new SystemMessage($content);
     }
 
     /**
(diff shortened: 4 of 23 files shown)


src/Message/Content/ImageUrlContent.php
```diff
@@ -0,0 +1,23 @@
...
+
+    /**
+     * @return array{
+     *     type: 'image_url',
+     *     image_url: array{url: string}
+     * }
+     */
+    public function jsonSerialize(): array
+    {
+        return ['type' => 'image_url', 'image_url' => ['url' => $this->imageUrl]];
```
OskarStark, line 21:
```suggestion
    public function __construct(public string $url)
    {
    }

    /**
     * @return array{
     *     type: 'image_url',
     *     image_url: array{url: string}
     * }
     */
    public function jsonSerialize(): array
    {
        return ['type' => 'image_url', 'image_url' => ['url' => $this->url]];
```

OskarStark:
Another approach
* #26

OskarStark:
> public static function ofUser(string|Content ...$content): UserMessage

👍

src/OpenAI/Response/Response.php
```diff
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace PhpLlm\LlmChain\Response;
+namespace PhpLlm\LlmChain\OpenAI\Response;
```
chr-hertel, line 5: was it intentional to move the class? in my understanding that leads to design violations in classes like the `ToolBox` or the `LanguageModel` interface

DZunke: No. It was not intentional. I do not know whta i have mesed up locally but surely in the end i will do a clean up. Was just a fast push to show a proposal and have a discussion about the initial topic. Sorry for the change!

DZunke:
> ```
> public static function ofUser(string|Content ...$content): UserMessage
> ```
> 
> but this would be the trade off for having a nicer API
> 
> WDYT?

I am absolute fine with keeping the Message as an initial entry point. But would you be fine with the idea the message so being a facade for specific implementations of the messages to get rid of the growing Message DTO? That's what i understood.

From my PoV we could so combine DX and keeping cleaner DTOs with a specific purpose. Also with the `MessageInterface` there would be the direct enablement for more use cases from other models. 

I already understood to have a look for the Claude implementation because the image reader looks a bit different in the payload there. Something to thing about when serializing the message 🤔 With my initial idea i would have the direct idea with having a Content Message in the GPT and another one in the Claude space. So the developer can decide based on the utilized model. But with the central entrypoint i see the problem to decide during rendering based on the language model how to render the array.

chr-hertel:
> But would you be fine with the idea the message so being a facade for specific implementations of the messages to get rid of the growing Message DTO? That's what i understood.

Yes!

> Something to thing about when serializing the message

so that's bringing us to the next item here in general - better to scope out tho for the sake of PR size:

ofc the interface enables users to define their own messages, yes. better DX would be limited to built-in ones via Message facade (can't believe i'm favoring this pattern here :laughing:). the json representation is model specific and therefore we might end up using the serializer for this later with the model and it's version as a serialization context - but step by step

OskarStark:
IMHO the problem (but I am a bit tired) is only because we are using jsonSerialize() inside the message now. If the message would just be a data holder, we could serialize (with Symfony serializer) the way we need it. OFC only if the message holds all necessary information, so `Content::image(string $url, string $filetype......)`

chr-hertel:
@OskarStark :100: same idea - but let's go with a follow up PR to switch to serializer for that process

src/Message/Content/ImageUrlContent.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpLlm\LlmChain\Message\Content;
+
+final readonly class ImageUrlContent implements ContentInterface
```
chr-hertel, line 7: what about just?
```suggestion
final readonly class Image implements Content
```

src/Message/UserMessage.php
```diff
@@ -0,0 +1,56 @@
...
+{
+    /**
+     * @var list<ImageUrlContent|string>
+     */
+    public array $images;
+
+    public function __construct(
+        public TextContent|string $content,
+        ImageUrlContent|string ...$images,
+    ) {
```
chr-hertel, line 20: i feel like this is a bit limited in comparison to the contract in the end which is just a list of content entries.

that's why i proposed `Message::ofUser(string|Content ...$content)` in the first place.
my thought was it's an array of `Content` but it would be nice to enable just a string as well.

so the logic could be the Message service method does an upcasting of a `string` entries of that `$content` array to `Text` instances, since this is the most prominent use-case. `Image` instances would be properly instantiated by the user themselves.

WDYT?

DZunke: Aaaah! Yeah ... ok ... thats fine! Will do it and adjust the tests. Had the upcasting now in the UserMessage Object but yeah, moving it a bit forward. Would then look like the following with images.

```
// Text as String
Message::ofUser('foo', new Image('http://bar'), new Image ('http://baz'));
// Result: Text, Image, Image

# Text as Content
Message::ofUser(new Text('foo'), new Image('http://bar'), new Image ('http://baz'));
// Result: Text, Image, Image

# All as String
Message::ofUser('foo', 'http://bar', 'http://baz');
// Result: Text, Image, Image

# Only first string can be text ... all others has to be images
Message::ofUser('foo', 'http://bar', new Image ('http://baz'));
// Result: Text, Image, Image
```

What do you think? A lot of casting but looks something consistent?

chr-hertel: Yeah, thought about the http prefix upcasting for Images but quite skeptical if safe to assume it's always an Image. I know for example that audio support was planned and could Imagine URL Support for that as well.
But we could just break the contract later on as well 👍

DZunke: Checking for a valid HTTP URL would not work. You could also embbed images. That is what i personally do. So base64 encoded images are always valid as a string. So the check would have to also check for data urls 🤔

chr-hertel: let's keep it simple for now. only upcast is `string` => `new Text` - if you want an image use `new Image`

fair?

DZunke: 😯  ... Ähm ... yeah .. simple ... already wrote the bit more complex code with tests. If you want to keep it simple i am open to remove the complexity from the code. Should i? 😉

src/Message/Message.php
```diff
@@ -28,9 +30,35 @@ public static function ofAssistant(?string $content = null, ?array $toolCalls =
...
+            if (0 === $index) {
+                $text = $entry;
+
+                if (\is_string($text)) {
+                    $text = new Text($entry);
+                }
+
+                if (!$text instanceof Text) {
+                    throw new \InvalidArgumentException('The first content piece has to be a string or Text part.');
+                }
```
chr-hertel, line 49: opinion or fact? the example of Oskar demonstrated that OpenAI is fine without a text first or did i get it wrong?

DZunke: Fact for the complex parsing as we discussed before. With all those castings. So in this case more of an opinion. Generally it would also work completely without text. The text is just giving instructions on how to interpret the images in the payload.

chr-hertel: i guess we had a misunderstanding than - sorry! now i get your previous comment ... one sec

src/Message/UserMessage.php
DZunke: So... the answer was yes: https://github.com/php-llm/llm-chain/pull/25#discussion_r1773830195 😜

chr-hertel: sorry, didn't catch that in time

src/Message/Message.php
DZunke: Now it is the simplified version. Mapping all strings to texts. Just had to put in the `UserMessage` that when only a single text is there it will utilize the simple format for serialization instead of the complex content format.

src/Message/UserMessage.php
```diff
@@ -0,0 +1,50 @@
...
+
+final readonly class UserMessage implements MessageInterface
+{
+    /**
+     * @var list<ContentInterface>
+     */
+    public array $content;
+
+    public function __construct(
+        ContentInterface ...$images,
```
OskarStark, line 19: Why is this variable named `$images`?

examples/image-describer.php
```diff
@@ -0,0 +1,27 @@
...
+$runtime = new OpenAI(HttpClient::create(), getenv('OPENAI_API_KEY'));
+$llm = new Gpt($runtime, Version::gpt4oMini());
+
+$chain = new Chain($llm);
+$messages = new MessageBag(
+    Message::forSystem('You are an image analyzer that looks to images like a comedian would like.'),
+    Message::ofUser(
+        'Describe the image as a comedian would do it.',
+        'https://upload.wikimedia.org/wikipedia/commons/thumb/3/31/Webysther_20160423_-_Elephpant.svg/350px-Webysther_20160423_-_Elephpant.svg.png',
+        'https://upload.wikimedia.org/wikipedia/commons/thumb/3/37/African_Bush_Elephant.jpg/320px-African_Bush_Elephant.jpg',
```
OskarStark, line 22: based on the code in this PR; this example looks wrong to me

DZunke: Why? Works fine and for me it looks good. Maybe grammar errors.

src/Message/UserMessage.php
DZunke: Because third time rebuilding. Thanks for pointing this out .

tests/Message/MessageBagTest.php
```diff
@@ -56,7 +59,11 @@ public function with(): void
 
         self::assertCount(3, $messageBag);
         self::assertCount(4, $newMessageBag);
-        self::assertSame('It is time to wake up.', $newMessageBag[3]->content);
+
+        $newMessageFromBag = $newMessageBag[3];
+        assert($newMessageFromBag instanceof AssistantMessage);
```
OskarStark, line 64: can we use the assertion method of PHPUnit for such things please?

examples/image-describer.php
OskarStark: Wouldn't we need to use `Image()` here or what am I missing?

src/Message/Content/Image.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpLlm\LlmChain\Message\Content;
+
+final readonly class Image implements ContentInterface
+{
+    public function __construct(public string $image)
```
OskarStark, line 9:
```suggestion
    public function __construct(public string $url)
```

And maybe also assert it starts with `http` in the actor? Or too much?

chr-hertel: too much, data urls are also valid

examples/image-describer.php
```diff
@@ -0,0 +1,27 @@
...
+$runtime = new OpenAI(HttpClient::create(), getenv('OPENAI_API_KEY'));
+$llm = new Gpt($runtime, Version::gpt4oMini());
+
+$chain = new Chain($llm);
+$messages = new MessageBag(
+    Message::forSystem('You are an image analyzer bot that helps identify the content of images.'),
+    Message::ofUser(
+        'Describe the images as a comedian would do it.',
+        'https://upload.wikimedia.org/wikipedia/commons/thumb/3/31/Webysther_20160423_-_Elephpant.svg/350px-Webysther_20160423_-_Elephpant.svg.png',
+        'https://upload.wikimedia.org/wikipedia/commons/thumb/3/37/African_Bush_Elephant.jpg/320px-African_Bush_Elephant.jpg',
```
chr-hertel, line 22: This is not sufficient as string anymore

```suggestion
        new Image('https://upload.wikimedia.org/wikipedia/commons/thumb/3/31/Webysther_20160423_-_Elephpant.svg/350px-Webysther_20160423_-_Elephpant.svg.png'),
        new Image('https://upload.wikimedia.org/wikipedia/commons/thumb/3/37/African_Bush_Elephant.jpg/320px-African_Bush_Elephant.jpg'),
```

examples/image-describer.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+use PhpLlm\LlmChain\Chain;
+use PhpLlm\LlmChain\Message\Message;
```
chr-hertel, line 4:
```suggestion
use PhpLlm\LlmChain\Chain;
use PhpLlm\LlmChain\Message\Content\Image;
use PhpLlm\LlmChain\Message\Message;
```

examples/image-describer.php
chr-hertel: might still work bc gpt knows those public urls, but the result of this is:

```json
    {
        "role": "system",
        "content": "You are an image analyzer bot that helps identify the content of images."
    },
    {
        "role": "user",
        "content": [
            {
                "type": "text",
                "text": "Describe the images as a comedian would do it."
            },
            {
                "type": "text",
                "text": "https:\/\/upload.wikimedia.org\/wikipedia\/commons\/thumb\/3\/31\/Webysther_20160423_-_Elephpant.svg\/350px-Webysther_20160423_-_Elephpant.svg.png"
            },
            {
                "type": "text",
                "text": "https:\/\/upload.wikimedia.org\/wikipedia\/commons\/thumb\/3\/37\/African_Bush_Elephant.jpg\/320px-African_Bush_Elephant.jpg"
            }
        ]
    }
```

src/Message/Content/Image.php
chr-hertel: e.g. something like this is also valid:
```json
        {
          "type": "image_url",
          "image_url": {
            "url": f"data:image/jpeg;base64,{base64_image}"
          }
        }
```

src/Message/UserMessage.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpLlm\LlmChain\Message;
+
+use PhpLlm\LlmChain\Message\Content\ContentInterface;
+use PhpLlm\LlmChain\Message\Content\Image;
+use PhpLlm\LlmChain\Message\Content\Text;
```
chr-hertel, line 9: this should be a pipeline finding ... :D

```suggestion
use PhpLlm\LlmChain\Message\Content\ContentInterface;
use PhpLlm\LlmChain\Message\Content\Text;
```

src/Message/UserMessage.php
```diff
@@ -0,0 +1,50 @@
...
+        $array = ['role' => Role::User];
+        if (1 === count($this->content) && $this->content[0] instanceof Text) {
+            $array['content'] = $this->content[0]->text;
+
+            return $array;
+        }
+
+        foreach ($this->content as $entry) {
+            $array['content'][] = $entry->jsonSerialize();
+        }
```
chr-hertel, line 46: should be sufficient to just call
```suggestion
        $array['content'] = $this->content;
```

tests/Message/MessageBagTest.php
chr-hertel: like
```suggestion
        self::assertInstanceOf(AssistantMessage::class, $messageFromBag);
```

tests/Message/MessageBagTest.php
```diff
@@ -73,7 +80,11 @@ public function merge(): void
         ));
 
         self::assertCount(4, $messageBag);
-        self::assertSame('It is time to wake up.', $messageBag[3]->content);
+
+        $messageFromBag = $messageBag[3];
+        assert($messageFromBag instanceof AssistantMessage);
```
chr-hertel, line 85:
```suggestion
        self::assertInstanceOf(AssistantMessage::class, $messageFromBag);
```

tests/Message/MessageBagTest.php
```diff
@@ -89,7 +100,11 @@ public function withoutSystemMessage(): void
 
         self::assertCount(3, $messageBag);
         self::assertCount(2, $newMessageBag);
-        self::assertSame('It is time to sleep.', $newMessageBag[0]->content);
+
+        $messageFromNewBag = $newMessageBag[0];
+        assert($messageFromNewBag instanceof AssistantMessage);
```
chr-hertel, line 105:
```suggestion
        self::assertInstanceOf(AssistantMessage::class, $messageFromBag);
```

tests/Message/MessageBagTest.php
```diff
@@ -105,7 +120,11 @@ public function prepend(): void
 
         self::assertCount(2, $messageBag);
         self::assertCount(3, $newMessageBag);
-        self::assertSame('My amazing system prompt.', $newMessageBag[0]->content);
+
+        $newMessageBagMessage = $newMessageBag[0];
+        assert($newMessageBagMessage instanceof SystemMessage);
```
chr-hertel, line 125:
```suggestion
        self::assertInstanceOf(SystemMessage::class, $messageFromBag);
```

chr-hertel:
Nice one, pretty much like that thing now :) left only a few minor things, but mostly as suggestion to just merge.

something's still not right with the pipeline, sorry for that

examples/image-describer.php
DZunke: Yeah! Thanks for the hint. Missed out to change the example after the last interface change.

examples/image-describer.php
DZunke: Yep. It is what Oskar already mentioned. Missed it after latest interface change.

src/Message/Content/Image.php
OskarStark, line 9: I would rename it to 
```suggestion
    /** @param $url An URL like http..... or a data url.... like ....
    public function __construct(public string $url)
```
WDYT?

src/Message/UserMessage.php
DZunke: Nice. Thanks. Did not worked with this JsonSerializable interface a lot in the past. Cool feature! 👍

DZunke: But! What to mention here. This does only work when utilizing `json_encode` it will not automatically parse it to arrays. So i have to adjust the tests to test encoded arrays instead of comparing the arrays. 

But for me it is fine. The code must surely not be different just to utilize the array instead of the json encoded array.

src/Message/Content/Image.php
DZunke: Had chosen the name because data url and http url are valid. But yeah. You are right. Sounds a bit better, at least together with a comment.

tests/Message/MessageBagTest.php
DZunke: Sure! Done.

chr-hertel (approved):
Thank you @DZunke!

chr-hertel:
one thing i didn't think about is the impact to using/handling/rendering the `MessageBag` in twig - potential follow up

for now i created https://github.com/php-llm/llm-chain-bundle/pull/22

and patched my example project like this:
image

PR #26, opened 2024-09-23
Support image url

OskarStark:
A simpler implementation of
* #25

cc @DZunke

### DIFF
--- a/tests/Message/MessageBagTest.php
+++ b/tests/Message/MessageBagTest.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace PhpLlm\LlmChain\Tests\Message;
 
+use PhpLlm\LlmChain\Message\ImageUrl;
 use PhpLlm\LlmChain\Message\Message;
 use PhpLlm\LlmChain\Message\MessageBag;
 use PHPUnit\Framework\Attributes\CoversClass;
@@ -115,6 +116,8 @@ final class MessageBagTest extends TestCase
             Message::forSystem('My amazing system prompt.'),
             Message::ofAssistant('It is time to sleep.'),
             Message::ofUser('Hello, world!'),
+            Message::ofUser('Analyze the following image:'),
+            Message::ofUser(new ImageUrl('http://example.com/image.jpg')),
         );
 
         $json = json_encode($messageBag);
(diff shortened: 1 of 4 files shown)


tests/Message/MessageBagTest.php
```diff
@@ -125,6 +128,10 @@ public function jsonSerialize(): void
                 ['role' => 'system', 'content' => 'My amazing system prompt.'],
                 ['role' => 'assistant', 'content' => 'It is time to sleep.'],
                 ['role' => 'user', 'content' => 'Hello, world!'],
+                ['role' => 'user', 'content' => 'Analyze the following image:'],
+                ['role' => 'user', 'content' => [
+                    ['type' => 'image_url', 'image_url' => ['url' => 'http://example.com/image.jpg']],
+                ]],
```
OskarStark, line 134: The only difference I propose is to use 2 different messages, would that be feasible for you @DZunke?

chr-hertel: with this we won't get the structure that is conform to the api specs:

```json
{
    "model": "gpt-4o-mini",
    "messages": [
      {
        "role": "user",
        "content": [
          {
            "type": "text",
            "text": "What’s in this image?"
          },
          {
            "type": "image_url",
            "image_url": {
              "url": "https://upload.wikimedia.org/wikipedia/commons/thumb/d/dd/Gfp-wisconsin-madison-the-nature-boardwalk.jpg/2560px-Gfp-wisconsin-madison-the-nature-boardwalk.jpg"
            }
          }
        ]
      }
    ],
    "max_tokens": 300
  }
```
see https://platform.openai.com/docs/guides/vision

chr-hertel: same for claude:

```json
{"role": "user", "content": [
  {
    "type": "image",
    "source": {
      "type": "base64",
      "media_type": "image/jpeg",
      "data": "/9j/4AAQSkZJRg...",
    }
  },
  {"type": "text", "text": "What is in this image?"}
]}
```
see https://docs.anthropic.com/en/api/messages

DZunke: Thanks for pointing it out with another design but yeah. As Christopher mentioned it would not fit the specification and also change the functionality to split it up because the text is the prompt how the image should be tackled as far as i understood it. A previous message would just be "context" and irrelevant in itself for the image processing.

OskarStark: It would not fulfill the spec, but yes it generates a valid answer as you can see in my screenshot above.

Issue #27, opened 2024-09-23
New name for `Runtime` concept


chr-hertel:
I am unsure about that `Runtime` naming and Oskar proposed `Platform`. I like that :)

This refactoring would also need to touch the `runtimes` config in llm-chain-bundle

chr-hertel:
Thanks @OskarStark for taking over :pray:

PR #34, opened 2024-09-25
Rename `Runtime` to `Platform`

OskarStark:
Refs
* #27 

### Needs (in order)
- [ ] `0.2` release of `php-llm/llm-chain`
- [ ] Merge of https://github.com/php-llm/llm-chain-bundle/pull/20
- [ ] `0.2` release of `php-llm/llm-chain-bundle`
- [ ] Closing of #27

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,19 +21,19 @@ composer require php-llm/llm-chain
 
 When using Symfony Framework, check out the integration bundle [php-llm/llm-chain-bundle](https://github.com/php-llm/llm-chain-bundle).
 
-Supported Models & Runtimes
----------------------------
+Supported Models & Platforms
+----------------------------
 
-Currently supported models and runtimes:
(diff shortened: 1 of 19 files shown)


chr-hertel:
Don't you think it's fine without tagging? don't care too much about semver at this point with 3 users and stating every tag as pre release :D
if you need it for your process, i'm def up for it - but i'm also lazy

chr-hertel (approved):
reviewed and tested - thank you :pray:

chr-hertel:
Tagged the version :+1:

OskarStark:
> Don't you think it's fine without tagging? don't care too much about semver at this point with 3 users and stating every tag as pre release :D

I agree, but a pipeline for the bundle with lowest and highest would fail otherwise without a new release, right?

PR #37, opened 2024-09-25
Add `topP` support

OskarStark:
see https://platform.openai.com/docs/api-reference/chat/create#chat-create-top_p

### DIFF
--- a/src/OpenAI/Model/Gpt.php
+++ b/src/OpenAI/Model/Gpt.php
@@ -11,6 +11,7 @@ use PhpLlm\LlmChain\OpenAI\Runtime;
 use PhpLlm\LlmChain\Response\Choice;
 use PhpLlm\LlmChain\Response\Response;
 use PhpLlm\LlmChain\Response\ToolCall;
+use Webmozart\Assert\Assert;
 
 final class Gpt implements LanguageModel
 {
@@ -18,8 +19,15 @@ final class Gpt implements LanguageModel
         private readonly Runtime $runtime,
         private ?Version $version = null,
         private readonly float $temperature = 1.0,
+        private readonly float $topP = 1.0,
     ) {
         $this->version ??= Version::gpt4o();
+
+        Assert::greaterThanEq($temperature, 0.01);
+        Assert::lessThanEq($temperature, 2);
+
+        Assert::greaterThanEq($topP, 0.01);
+        Assert::lessThanEq($topP, 1);
     }
 
     public function call(MessageBag $messages, array $options = []): Response
@@ -27,6 +35,7 @@ final class Gpt implements LanguageModel
         $body = [
             'model' => $this->version->name,
             'temperature' => $this->temperature,
+            'top_p' => $this->topP,
             'messages' => $messages,
         ];
 


src/OpenAI/Model/Gpt.php
```diff
@@ -11,22 +11,31 @@
...
 use PhpLlm\LlmChain\Response\ToolCall;
+use Webmozart\Assert\Assert;
 
 final class Gpt implements LanguageModel
 {
     public function __construct(
         private readonly Runtime $runtime,
         private ?Version $version = null,
         private readonly float $temperature = 1.0,
+        private readonly float $topP = 1.0,
```
chr-hertel, line 22: What do you think in general of having those parameters here? Struggle a bit with my initial decision - should maybe only the part of the `$options` of `call(..)` method?

chr-hertel (commented):
Only a general question - no hard feelings.

src/OpenAI/Model/Gpt.php
DZunke: I think for default settings for fine tuning this position is fitting. But maybe all settings available should be put to a config object? 

When i have a look to the OpenAI Playground ... there are more settings beside temperature and topP for every request. So this would stack up to this amount.

image

 Maybe this could also be utilized as argument by the user?

chr-hertel: maintaining a config object would be hard - the api contract is growing and depends on versions and platform. we could tackle that of course with having designated extension points, but having an array is way more flexible for now :speak_no_evil: :laughing:

OskarStark: We can go with `$options` array then, but we couldn't validate them, just pass them to the model -> fire and forget... 🤷‍♂️

chr-hertel: We need way better error handling tho for http calls. Was quite lazy with that ... 👼

src/OpenAI/Model/Gpt.php
```diff
@@ -11,22 +11,31 @@
...
     ) {
         $this->version ??= Version::gpt4o();
+
+        Assert::greaterThanEq($temperature, 0.01);
```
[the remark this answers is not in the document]
OskarStark: Based on my research it needs to be at least `0.01`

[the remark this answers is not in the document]
chr-hertel: Research meaning usage? Pretty sure i used 0 before but can check tomorrow or later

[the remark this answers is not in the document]
OskarStark: CleanShot 2024-09-25 at 22 09 11@2x

Yes you are right, based on the docs, however, when I played with the assistants:
CleanShot 2024-09-25 at 22 10 33@2x

You cannot go below `0.01`

OskarStark:
Superseded by 
* #41

src/OpenAI/Model/Gpt.php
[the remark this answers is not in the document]
chr-hertel: UI says no, API says yes

image

image

PR #39, opened 2024-09-25
Split pipeline

OskarStark:
No need to run cs-fixer, PHPStan etc. so often with different PHP versions

Needs
* #40


OskarStark:
This PR needs to be merged with admin rights and then afterwards, you need to adjust the branch protection please. Thanks

OskarStark:
Rebased

PR #42, opened 2024-09-25
Add MongoDB vector store

OskarStark:
I need to test this in my project first

cc @alcaeus

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ Supported Stores
 
 * [x] [ChromaDB](https://trychroma.com)
 * [x] [Azure AI Search](https://azure.microsoft.com/en-us/products/ai-services/ai-search)
+* [x] [MongoDB Atlas Search](https://mongodb.com/products/platform/atlas-vector-search)
 * [ ] [Pinecone](https://pinecone.io)
 
 Provided Tools
--- a/src/Store/MongoDB/Store.php
+++ b/src/Store/MongoDB/Store.php
@@ -1,0 +1,152 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpLlm\LlmChain\Store\MongoDB;
+
+use MongoDB\BSON\Binary;
+use MongoDB\Client;
+use MongoDB\Collection;
+use PhpLlm\LlmChain\Document\Document;
+use PhpLlm\LlmChain\Document\Metadata;
+use PhpLlm\LlmChain\Document\Vector;
+use PhpLlm\LlmChain\Store\VectorStoreInterface;
+use Psr\Log\LoggerInterface;
+use Symfony\Component\Uid\Uuid;
+
+/**
+ * @see https://www.mongodb.com/docs/atlas/atlas-vector-search/vector-search-overview/
+ *
+ * For this store you need to create a separate MongoDB Atlas Search index.
+ * The index needs to be created with the following settings:
+ * {
+ *     "fields": [
+ *         {
+ *             "numDimensions": 1536,
+ *             "path": "vector",
+ *             "similarity": "euclidean",
+ *             "type": "vector"
+ *         }
+ *     ]
+ * }
+ *
+ * Note, that the `path` key needs to match the $vectorFieldName.
+ *
+ * For the `similarity` key you can choose between `euclidean`, `cosine` and `dotProduct`.
+ * {@see https://www.mongodb.com/docs/atlas/atlas-search/field-types/knn-vector/#define-the-index-for-the-fts-field-type-type}
+ *
+ * @author Oskar Stark <oskarstark@googlemail.com>
+ */
+final readonly class Store implements VectorStoreInterface
+{
+    /**
+     * @param string $databaseName    The name of the database
+     * @param string $collectionName  The name of the collection
+     * @param string $indexName       The name of the Atlas Search index
+     * @param string $vectorFieldName The name of the field int the index that contains the vector
+     * @param bool   $bulkWrite       Use bulk write operations
+     */
+    public function __construct(
+        private Client $client,
+        private LoggerInterface $logger,
+        private string $databaseName,
+        private string $collectionName,
+        private string $indexName,
+        private string $vectorFieldName = 'vector',
+        private bool $bulkWrite = false,
+    ) {
+    }
+
+    public function addDocument(Document $document): void
+    {
+        $this->addDocuments([$document]);
+    }
+
+    public function addDocuments(array $documents): void
+    {
+        $operations = [];
+
+        foreach ($documents as $document) {
+            if (!$document->hasVector()) {
+                $this->logger->warning('Document {id} does not have a vector', ['id' => $document->id]);
+            }
+
+            $operation = [
+                ['_id' => $this->toBinary($document->id)], // we use binary for the id, because of storage efficiency
+                array_filter([
+                    'metadata' => $document->metadata,
+                    $this->vectorFieldName => $document->vector->getData(),
+                    'text' => $document->text,
+                ]),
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -45,6 +45,7 @@ Supported Stores
 
 * [x] Chroma DB
 * [x] Azure AI Search
+* [x] MongoDB Atlas Search
```
chr-hertel, line 48: needs a link now :D

README.md
```diff
@@ -43,8 +43,10 @@ Planned Models & Platforms (not implemented yet):
 Supported Stores
 ----------------
 
+<<<<<<< HEAD
```
alcaeus, line 46: You might want to remove the conflict marker

src/Store/MongoDB/Store.php
```diff
@@ -0,0 +1,133 @@
...
+use PhpLlm\LlmChain\Document\Metadata;
+use PhpLlm\LlmChain\Document\Vector;
+use PhpLlm\LlmChain\Store\VectorStoreInterface;
+use Psr\Log\LoggerInterface;
+use Symfony\Component\Uid\Uuid;
+
+/**
+ * @see https://www.mongodb.com/docs/atlas/atlas-vector-search/vector-search-overview/
+ *
+ * For this store you need t create a separate MongoDB Atlas Search index.
```
alcaeus, line 18:
```suggestion
 * For this store you need to create a separate MongoDB Atlas Search index.
```

src/Store/MongoDB/Store.php
```diff
@@ -0,0 +1,133 @@
...
+
+        foreach ($documents as $document) {
+            if (!$document->hasVector()) {
+                $this->logger->warning('Document {id} does not have a vector', ['id' => $document->id]);
+            }
+
+            $id = $document->id->toRfc4122();
+
+            $collection->replaceOne(
+                ['id' => $id],
```
alcaeus, line 72: I noticed that there's no information on an index being required on this field. You should consider using the default `_id` field for this to leverage the index. Note that the identifier does not need to be an ObjectId, but can be any acceptable type for MongoDB fields.

src/Store/MongoDB/Store.php
```diff
@@ -0,0 +1,133 @@
...
+                    '_id' => 0,
+                    'id' => 1,
+                    'text' => 1,
+                    'metadata' => 1,
+                    $this->vectorFieldName => 1,
+                    'score' => [
+                        '$meta' => 'vectorSearchScore',
+                    ],
+                ],
+            ],
```
alcaeus, line 117: Unless you want to specifically exclude other fields than ID, I recommend not using `$project` but rather `$addFields`. Assuming you're using the default `_id` field for the identifier, the following stage would accomplish what you want:
```suggestion
            [
                '$addFields' => [
                    'id' => '$_id',
                    'score' => [ '$meta' => 'vectorSearchScore' ],
                ],
            ],
```

src/Store/MongoDB/Store.php
```diff
@@ -0,0 +1,133 @@
...
+        ]);
+
+        $documents = [];
+        foreach ($results as $result) {
+            $result = $result->getArrayCopy();
+
+            $documents[] = Document::fromVector(
+                Vector::create1536($result[$this->vectorFieldName]),
+                Uuid::fromString($result['id']),
```
alcaeus, line 126: Noticed that you're storing a UUID as the identifier. Consider storing this as a [binary field](https://www.php.net/manual/en/class.mongodb-bson-binary.php#mongodb-bson-binary.constants.type-uuid) for better storage efficiency

README.md
OskarStark: Jepp

src/Store/MongoDB/Store.php
OskarStark: Thanks

OskarStark:
Thanks for the review @alcaeus

src/Store/MongoDB/Store.php
OskarStark: You mean like this?

src/Store/MongoDB/Store.php
```diff
@@ -0,0 +1,128 @@
...
+ * {@see https://www.mongodb.com/docs/atlas/atlas-search/field-types/knn-vector/#define-the-index-for-the-fts-field-type-type}
+ *
+ * @author Oskar Stark <oskarstark@googlemail.com>
+ */
+final readonly class Store implements VectorStoreInterface
+{
+    /**
+     * @param string $databaseName    The name of the database
+     * @param string $collectionName  The name of the collection
+     * @param string $indexName       The name of the index
```
OskarStark, line 43:
```suggestion
     * @param string $indexName       The name of the Atlas Search index
```

src/Store/MongoDB/Store.php
alcaeus: Yep, better 👍

src/Store/MongoDB/Store.php
alcaeus: After you made these changes, I looked for my previous work with the BSON binary type and Uuid:
```php
use MongoDB\BSON\Binary;
use Symfony\Component\Uid\Uuid;

function toBinary(Uuid $uuid): Binary
{
    return new Binary($uuid->toBinary(), Binary::TYPE_UUID);
}

function toUuid(Binary $binary): Binary
{
    return Uuid::fromString($binary->getData());
}
```

OskarStark: done

alcaeus (approved):

src/Store/MongoDB/Store.php
```diff
@@ -0,0 +1,163 @@
...
+        private bool $bulkWrite = false,
+    ) {
+        $this->client = new Client(
+            uri: $uri,
+            driverOptions: [
+                'driver' => [
+                    'name' => 'php-llm/llm-chain',
+                ],
+            ],
+        );
```
chr-hertel, line 68: Had the same issue with the third party client at https://github.com/php-llm/llm-chain/blob/main/src/Store/ChromaDB/Store.php, but decided for injection instead of instatiation because the client is a service in my perspective. but i ended up wiring that client myself.

what do you think about a factory approach instead?

OskarStark: Sure I can go back to client injection.

Can you elaborate on the factory approach? How can that help here?

chr-hertel: the factory is basically an opinionated, but optional way to construct that store, which could be used by an implementation like our bundle, but this lib could be used in a different context where the client lifecycle is already handled differently and therefore injection would be more flexible.

chr-hertel: or maybe even the Client? :thinking:

OskarStark: Let's go with the client for know, because when you install the Symfony mongodb bundle you are good to go ✌🏻

chr-hertel: isn't the mongodb bundle exactly an argument for injecting the Client?

so i would configure my client using that bundle, e.g.
```
mongodb:
  clients:
    default:
      uri: '%env(MONGODB_URI)%'
      default_database: #...
      uri_options: #...
      driver_options: #...
```

and than i would expect that this is sufficient to use that client as a vector store or do i get something wrong?

chr-hertel (approved):
anyhow, fine by me, your use case for now and awesome to have the support in :+1: :sunglasses:

src/Store/MongoDB/Store.php
OskarStark: > isn't the mongodb bundle exactly an argument for injecting the Client?

Yes

PR #47, opened 2024-09-25
feat: introducing chain processors

chr-hertel:
Idea to decouple functionality and extension points for chain - not working yet though

WDYT?


OskarStark:
Sounds like a nice idea!

chr-hertel:
works now, but need to think about it ...

src/OpenAI/Model/Gpt.php
```diff
@@ -40,6 +40,11 @@ public function supportsToolCalling(): bool
         return true;
     }
 
+    public function supportsImageInput(): bool
+    {
+        return $this->version->supportImageInput;
+    }
```
OskarStark, line 46: The logic which belongs to image support should be extracted to a separate PR to avoid noise around the new chain processor feature

chr-hertel: True, will do so tomorrow

PR #48, opened 2024-09-25
Add header to files

OskarStark:


chr-hertel:
i a) don't feel comfortable with that and b) don't that's necessary - but appreciated, hope that's alright :pray:

PR #51, opened 2024-09-26
Initialize store

OskarStark:
### Needs
- [x] mongodb 1.20
- [x] https://github.com/shivammathur/setup-php/issues/868

Should be merged after
* #61

### DIFF
--- a/.env
+++ b/.env
@@ -18,5 +18,5 @@ AZURE_OPENAI_KEY=
 # For using SerpApi (tool)
 SERP_API_KEY=
 
-# For using MongoDB (store)
+# For using MongoDB Atlas (store)
 MONGODB_URI=
--- a/.env
+++ b/.env
@@ -17,3 +17,6 @@ AZURE_OPENAI_KEY=
 
 # For using SerpApi (tool)
 SERP_API_KEY=
+
+# For using MongoDB (store)
+MONGODB_URI=
--- a/composer.json
+++ b/composer.json
@@ -37,6 +37,9 @@
         "symfony/dotenv": "^6.4 || ^7.1",
         "symfony/var-dumper": "^6.4 || ^7.1"
     },
+    "conflict": {
+            "mongodb/mongodb": "<1.20"
+    },
     "suggest": {
         "codewithkyrian/chromadb-php": "For using the ChromaDB as retrieval vector store.",
         "mongodb/mongodb": "For using MongoDB Atlas as retrieval vector store.",
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
         "symfony/var-dumper": "^6.4 || ^7.1"
     },
     "conflict": {
-            "mongodb/mongodb": "<1.20"
+        "mongodb/mongodb": "<1.20"
     },
     "suggest": {
         "codewithkyrian/chromadb-php": "For using the ChromaDB as retrieval vector store.",
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
     },
     "require-dev": {
         "codewithkyrian/chromadb-php": "^0.2.1",
-        "mongodb/mongodb": "^1.19",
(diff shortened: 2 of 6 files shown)


chr-hertel:
idea is similar to messenger setup? i like :+1: - not sure where to use it atm tho

src/Store/MongoDB/Store.php
```diff
@@ -57,6 +58,26 @@ public function __construct(
...
+                        'type' => 'vector',
+                    ],
+                ],
+            ],
+            options: [
+                'name' => $this->indexName,
+                'type' => 'vectorSearch',
+            ],
+        );
+    }
```
OskarStark, line 79: @alcaeus, this looks like mentioned in:
https://www.mongodb.com/docs/atlas/atlas-vector-search/vector-search-type/#replace-the-following-values-and-save-the-file.

But I receive:
CleanShot 2024-09-26 at 23 04 03@2x

Any clue?

OskarStark: It works now with mongodb/mongodb 1.20

@alcaeus any remarks?

.env
```diff
@@ -14,3 +14,6 @@ AZURE_OPENAI_KEY=
 
 # For using SerpApi (tool)
 SERP_API_KEY=
+
+# For using MongoDB (store)
```
OskarStark, line 18:
```suggestion
# For using MongoDB Atlas (store)
```

composer.json
```diff
@@ -37,6 +37,9 @@
         "symfony/dotenv": "^6.4 || ^7.1",
         "symfony/var-dumper": "^6.4 || ^7.1"
     },
+    "conflict": {
+        "mongodb/mongodb": "<1.20"
+    },
```
chr-hertel, line 42: this is not needed if the version constraint is hire anyways - composer would never install something below 1.20, right?

src/Store/InitializableStoreInterface.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpLlm\LlmChain\Store;
+
+interface InitializableStoreInterface extends StoreInterface
+{
+    /**
+     * @param array<mixed> $options
```
chr-hertel, line 10:
```suggestion
     * @param array<string, mixed> $options
```

src/Store/MongoDB/Store.php
```diff
@@ -135,6 +138,37 @@ public function query(Vector $vector, array $options = []): array
...
+     * @param array{fields?: array<mixed>} $options
+     */
+    public function initialize(array $options = []): void
+    {
+        try {
+            if ([] !== $options && !array_key_exists('fields', $options)) {
+                throw new \InvalidArgumentException('The only supported option is "fields"');
+            }
+
+            $this->getCollection()->createSearchIndex(
```
chr-hertel, line 151: would move the guard close out of the try catch block, since the intention is limited to `createSearchIndex`

```suggestion
        if ([] !== $options && !array_key_exists('fields', $options)) {
            throw new \InvalidArgumentException('The only supported option is "fields"');
        }

        try {
            $this->getCollection()->createSearchIndex(
```

src/Store/InitializableStoreInterface.php
OskarStark: I thought about this, but another store could be different, no?

composer.json
OskarStark: If you install it without require dev, nothing prevents you from installing 1.19 IMHO

chr-hertel: True thing, meaning as consequences we would need that for all suggestions?

And on the other hand, the dependency is optional - so a hard conflict i also a lie. Imagine using the lib without mongodb store but having mongodb installed for other parts of the app ...

chr-hertel: Nvm, getting into too much of theory

chr-hertel (approved):

PR #55, opened 2024-09-26
feat: add pinecone store

chr-hertel:


src/Store/MongoDB/Store.php
```diff
@@ -126,7 +126,7 @@ public function query(Vector $vector, array $options = []): array
 
         foreach ($results as $result) {
             $documents[] = Document::fromVector(
-                Vector::create1536($result[$this->vectorFieldName]),
```
OskarStark, line 129: So the static one is still needed?

OskarStark:
Nice one

src/Store/MongoDB/Store.php
chr-hertel: Nope, just thought it is an unnecessary Detail and limitation here

PR #57, opened 2024-09-27
Remove method `Vector::create1536`

OskarStark:


chr-hertel (approved):
once again yagni strikes back - well intended, not needed - thanks for cleaning up

PR #58, opened 2024-09-27
feat: support streaming of responses

chr-hertel:
tough nut and still not happy with this implementation - feedback welcome

https://github.com/user-attachments/assets/c88c5bc2-0c93-401f-b95f-a659ebb35035


chr-hertel:
on hold for refactoring of Message, Response and Document refactoring.

chr-hertel:
def not happy with the direction of `Response` obj as well, but would rather continue to add feature and shift refactoring to later with more implementations and knowledge

PR #74, opened 2024-10-01
Add support for response format in tools

### LINKED ISSUE #49: Feature Idea: Structured Output for Tools
### Description:

I’d like to suggest a feature that enables structured output for tools when interacting with the model. Currently, the model decides how to format and display the results, which may not always fit specific needs. This feature would allow users to specify the desired output format for their tools.

### Example:

In my case, I have a tool that fetches products from my API and presents them to the user. However, I don’t want the model to decide the structure of the output. Instead, I’ve specified a JSON schema that dictates the format of the response.

For instance, the result would look something like this:
```json
{
  "type": "products",
  "products": [
    {
      "name": "Product 1",
      "url": "https://example.com/product1",
      "price": 29.99
    },
    {
      "name": "Product 2",
      "url": "https://example.com/product2",
      "price": 49.99
    }
  ]
}
```

With this response, my frontend knows which renderer to use.

### Expected Behavior:

- The model should follow the given JSON schema when returning the result.
- The user should have the ability to predefine how the output is structured, without the model altering it.

This would improve consistency and allow for more flexible tool integrations.

OskarStark:
closes #49

I avoided to add logic for closures for now

### DIFF
--- a/examples/toolbox-clock.php
+++ b/examples/toolbox-clock.php
@@ -33,4 +33,5 @@ $chain = new Chain($llm, [$processor], [$processor]);
 $messages = new MessageBag(Message::ofUser('What date and time is it?'));
 $response = $chain->call($messages);
 
-echo $response->getContent().PHP_EOL;
+// Clock tool forced a structured output via #[AsTool] attribute
+dump(json_decode($response->getContent(), true));
--- a/src/ToolBox/Attribute/AsTool.php
+++ b/src/ToolBox/Attribute/AsTool.php
@@ -4,13 +4,26 @@ declare(strict_types=1);
 
 namespace PhpLlm\LlmChain\ToolBox\Attribute;
 
+use PhpLlm\LlmChain\Exception\InvalidArgumentException;
+
 #[\Attribute(\Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]
 final readonly class AsTool
 {
+    /**
+     * @param class-string|array<mixed>|null $responseFormat
+     */
     public function __construct(
         public string $name,
         public string $description,
         public string $method = '__invoke',
+        public string|array|null $responseFormat = null,
     ) {
+        if (is_string($responseFormat) && !class_exists($responseFormat)) {
+            throw new InvalidArgumentException('The response format must be a class name.');
+        }
+
+        if (is_array($responseFormat) && [] === $responseFormat) {
+            throw new InvalidArgumentException('The response format must not be an empty array.');
+        }
     }
 }
--- a/src/ToolBox/ChainProcessor.php
+++ b/src/ToolBox/ChainProcessor.php
@@ -38,16 +38,35 @@ final class ChainProcessor implements InputProcessor, OutputProcessor, ChainAwar
     {
         $messages = clone $output->messages;
 
+        $options = $output->options;
+
         while ($output->response instanceof ToolCallResponse) {
             $toolCalls = $output->response->getContent();
             $messages[] = Message::ofAssistant(toolCalls: $toolCalls);
 
             foreach ($toolCalls as $toolCall) {
+                foreach ($this->toolBox->getMap() as $metadata) {
+                    if ($metadata->name === $toolCall->name
+                        && null !== $metadata->responseFormat
+                    ) {
+                        $options = array_merge($options, [
+                            'response_format' => [
+                                'type' => 'json_schema',
+                                'json_schema' => [
+                                    'schema' => $metadata->responseFormat,
+                                    'name' => $metadata->name,
+                                ],
+                            ],
+                        ]);
+                        break;
+                    }
+                }
+
                 $result = $this->toolBox->execute($toolCall);
                 $messages[] = Message::ofToolCall($toolCall, $result);
             }
 
-            $output->response = $this->chain->call($messages, $output->options);
+            $output->response = $this->chain->call($messages, $options);
         }
     }
 }
--- a/src/ToolBox/Metadata.php
+++ b/src/ToolBox/Metadata.php
@@ -11,6 +11,7 @@ final readonly class Metadata implements \JsonSerializable
 {
     /**
      * @param ParameterDefinition|null $parameters
+     * @param array<mixed>|null        $responseFormat
      */
     public function __construct(
         public string $className,
@@ -18,6 +19,7 @@ final readonly class Metadata implements \JsonSerializable
         public string $description,
         public string $method,
         public ?array $parameters,
+        public ?array $responseFormat = null,
     ) {
     }
 
@@ -27,7 +29,8 @@ final readonly class Metadata implements \JsonSerializable
      *     function: array{
      *         name: string,
      *         description: string,
-     *         parameters?: ParameterDefinition
+     *         parameters?: ParameterDefinition,
+     *         responseFormat?: array<mixed>|null
      *     }
      * }
      */
@@ -40,6 +43,10 @@ final readonly class Metadata implements \JsonSerializable
 
         if (isset($this->parameters)) {
             $function['parameters'] = $this->parameters;
+        }
+
+        if (null !== $this->responseFormat) {
+            $function['responseFormat'] = $this->responseFormat;
         }
 
         return [
--- a/src/ToolBox/ToolAnalyzer.php
+++ b/src/ToolBox/ToolAnalyzer.php
@@ -5,12 +5,14 @@ declare(strict_types=1);
 namespace PhpLlm\LlmChain\ToolBox;
 
 use PhpLlm\LlmChain\Exception\InvalidToolImplementation;
+use PhpLlm\LlmChain\StructuredOutput\SchemaFactory;
 use PhpLlm\LlmChain\ToolBox\Attribute\AsTool;
 
 final readonly class ToolAnalyzer
 {
     public function __construct(
         private ParameterAnalyzer $parameterAnalyzer = new ParameterAnalyzer(),
+        private SchemaFactory $schemaFactory = new SchemaFactory(),
     ) {
     }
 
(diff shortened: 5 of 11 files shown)


src/StructuredOutput/ChainProcessor.php
```diff
@@ -16,7 +16,7 @@ final class ChainProcessor implements InputProcessor, OutputProcessor
     private string $outputStructure;
 
     public function __construct(
-        private readonly ResponseFormatFactory $responseFormatFactory,
+        private readonly ResponseFormatFactoryInterface $responseFormatFactory,
```
OskarStark, line 19: I added an interface to be able to test the ChainProcessor, maybe I am wrong, can you have a look?

src/ToolBox/AsTool.php
OskarStark, line 1: I think this class looks ok

src/ToolBox/ChainProcessor.php
```diff
@@ -41,6 +41,8 @@ public function processOutput(Output $output): mixed
                 $messages[] = Message::ofToolCall($toolCall, $result);
             }
 
+            // we need to check if the tool has a response format defined and apply that to the following call
```
OskarStark, line 44: Todo

src/ToolBox/Metadata.php
```diff
@@ -11,13 +11,15 @@
 {
     /**
      * @param ParameterDefinition|null $parameters
+     * @param array<mixed>|null        $responseFormat
```
OskarStark, line 14: The resolving of the responseFormat should be done before, so that we just have to deal with null (no response format) or a valid response format (array) here in the `Metadata` class, I hope you agree on that

src/ToolBox/Metadata.php
```diff
@@ -27,7 +29,8 @@ public function __construct(
      *     function: array{
      *         name: string,
      *         description: string,
-     *         parameters?: ParameterDefinition
+     *         parameters?: ParameterDefinition,
+     *         responseFormat?: array<mixed>|string|null
```
OskarStark, line 33:
```suggestion
     *         responseFormat?: array<mixed>|null
```

src/ToolBox/ToolAnalyzer.php
```diff
@@ -5,12 +5,17 @@
...
 use PhpLlm\LlmChain\Exception\InvalidToolImplementation;
+use PhpLlm\LlmChain\StructuredOutput\SchemaFactory;
 
 final readonly class ToolAnalyzer
 {
+    private SchemaFactory $schemaFactory;
+
     public function __construct(
         private ParameterAnalyzer $parameterAnalyzer = new ParameterAnalyzer(),
+        ?SchemaFactory $schemaFactory = null,
```
OskarStark, line 16: We would need the `SchemaFactory` here now, because of resolving the ResponseFormat before creating the `Metadata` object, good to go? Concerns?

tests/ToolBox/ChainProcessorTest.php
```diff
@@ -0,0 +1,115 @@
...
+        $llm = $this->createMock(LanguageModel::class);
+        $llm->method('supportsStructuredOutput')->willReturn(false);
+
+        $input = new Input($llm, new MessageBag(), ['output_structure' => 'SomeStructure']);
+
+        $chainProcessor->processInput($input);
+    }
+
+    #[Test]
+    public function testProcessOutputWithResponseFormat(): void
```
OskarStark, line 74: @chr-hertel need help on this:
```
There was 1 failure:

1) PhpLlm\LlmChain\Tests\ToolBox\ChainProcessorTest::testProcessOutputWithResponseFormat
null does not match expected type "object".

/Users/oskar.stark/dev/php-llm/llm-chain/tests/ToolBox/ChainProcessorTest.php:95
```

src/StructuredOutput/ChainProcessor.php
OskarStark: If you agree, I will move this to an extra PR

chr-hertel: gave it a look and another thought, go ahead - best option as well. if someone wants to fiddle around with the ResponseFactory i would rather enforce them to use composition over inheritance . classic :+1:

OskarStark: Done in 
* #93

tests/ToolBox/ChainProcessorTest.php
OskarStark: Moved this to a separate PR in
* https://github.com/php-llm/llm-chain/pull/97

src/ToolBox/ToolAnalyzer.php
chr-hertel: that's the reality if we want to combine those features, true :+1:

src/ToolBox/Attribute/AsTool.php
```diff
@@ -4,13 +4,26 @@
...
         public string $method = '__invoke',
+        public string|array|null $responseFormat = null,
     ) {
+        if (is_string($responseFormat) && !class_exists($responseFormat)) {
+            throw new InvalidArgumentException('The response format must be a class name.');
+        }
+
+        if (is_array($responseFormat) && [] === $responseFormat) {
+            throw new InvalidArgumentException('The response format must not be an empty array.');
+        }
```
chr-hertel, line 27: this is your major use case, right? or the other one as well?

OskarStark: You mean array or class-string? I only use arrays in my project

OskarStark:
Rebased

chr-hertel:
does it make sense to bring in an example here?

OskarStark:
Yes, 👍🏻

examples/toolbox-clock.php
```diff
@@ -33,4 +33,5 @@
 $messages = new MessageBag(Message::ofUser('What date and time is it?'));
 $response = $chain->call($messages);
 
-echo $response.PHP_EOL;
+// Clock tool forced a structured output
+dump(json_decode($response, true));
```
OskarStark, line 37: CleanShot 2024-10-04 at 22 35 10@2x

This is now working @chr-hertel, but maybe you can help me polish the part in `ChainProcessor`. I am also not that deep in the ChainProcessor thing and I really don't get the difference between response_format and output_structure there. Thanks

src/ToolBox/ChainProcessor.php
```diff
@@ -33,15 +33,34 @@ public function processOutput(Output $output): mixed
...
+                        break;
+                    }
+                }
+
                 $result = $this->toolBox->execute($toolCall);
                 $messages[] = Message::ofToolCall($toolCall, $result);
             }
 
-            $response = $output->llm->call($messages, $output->options);
+            $response = $output->llm->call($messages, $options);
```
chr-hertel, line 63: this cannot work like this. or maybe it does with the example, but

a) multiple tool calls will interfere with each other
b) the original output format get overwritten

chr-hertel: just realized why my assumption was wrong. the processor calls the llm not the chain - this is what makes it hard.

chr-hertel:
please have a look at #117 and in the end #118 if they provide what you were looking for

chr-hertel:
not sure this is still needed after #118 - guess the format is tackled and additional messages or option changes should be possible with custom processor - or do i miss something?

OskarStark:
> not sure this is still needed after https://github.com/php-llm/llm-chain/pull/118 - guess the format is tackled and additional messages or option changes should be possible with custom processor - or do i miss something?

What do you mean by "is not needed"? The change in the chain processor?


Feel free to overtake the PR, I "steh ein bisschen auf dem Schlauch" 😄 

Extra messages can be ignored for now, but I think yes, they are doable by your other PR, showcasing the `TestProcessor`

examples/toolbox-clock.php
```diff
@@ -33,4 +33,5 @@
 $messages = new MessageBag(Message::ofUser('What date and time is it?'));
 $response = $chain->call($messages);
 
-echo $response->getContent().PHP_EOL;
+// Clock tool forced a structured output
```
OskarStark, line 36:
```suggestion
// Clock tool forced a structured output via #[AsTool] attribute
```

OskarStark:
So the current state of my PR is still working after a rebase and provides me with the correct result. However, after removing my logic from the `ChainProcessor` it doesn't work anymore.

I would really appreciate if you could take over this PR, as I am bit lost here 😄 Thanks! 🙏

chr-hertel:
Let's have a call maybe friday or sth when I'm back home

OskarStark:
Hey, my use case is:

In my function I call an API and I get an array with products like:
```json
[
  {
    "name": "foo",
    "url": "/foo",    
  },
  {
    "name": "bar",
    "url": "/bar",    
  }
]
```

but I want to have dedicated response format, so what I do is, I just give it back as the response to the function call and in the same time I force a response format:
```json
[
  {
    "product_name": string,
    "target_url": string,    
  },
]
```

So I don't want to do the mapping myself, but the LLM should do it for me.
Again, but only if the answer is from THIS function.

I hope it is clear now

chr-hertel:
@OskarStark The issue is that function calls are handled as collection. there can be multiple of them with one model response and this makes it hard to "understand" which response_format we should use.

OskarStark:
Where you able to mange 2 function calls at once ?

DZunke:
Ah ok. So kind of what i have done, as i want to customize tool settings like description for the function and description of methods at runtime from users application settings, is to create my own toolbox with the help of the toolbox interface and overwrite the Metadata that are responded by the `getMap` functionality. So the tool box itself is already customizable for own purposes to give back own interpretation of metadata objects.

What i take from your answer and code change is that you want to reach a changed tools mapping from metadata objects to array in some cases? 

What my first simple idea was to extend the toolbox itself to build the mapping for the platforms. This way it could also be possible to have totally different tool definitions beside the attributes that are mapped to metadata objects. This would maybe also shrink the need with all the schema definitions combined within the classes with the simple string behavor as the metadata objects could be left out as a contract between platform and toolbox. 

I would understand this as a possibility to have a diverse set of custom interpreations available. Also to build complex json schema communication within the tools with own decisions based on the own use case instead of having the library to decide how to merge those complex JSON Schema together to a valid thingy. 

Maybe it should also be possible to have and extended tool box interface with methods for "render map as information which tools are available" and "optionaly render map for tools execution" as Input and Output in ChainProcessor seems target to be different? 

My use case is also currently working with the interface in place, as i mentioned, but i had those thoughts in general for an improvement after i had a revisit to this pull request.

What do you think? Maybe i am totally wrong as i am not utilizing the json schema approach and never tried the structured output in LLMs. Just wanted to bring this a bit further with sharing my thoughts.

chr-hertel:
@OskarStark this one is obsolete, isn't it?

PR #75, opened 2024-10-01
Add ability to describe tool parameters via `#[ToolParameter]` attribute

OskarStark:
### Todo

validate, that this is not allowed:
```php
#[ToolParameter(const: 'my-string')]
int $number
```

### DIFF
--- a/src/ToolBox/ParameterAnalyzer.php
+++ b/src/ToolBox/ParameterAnalyzer.php
@@ -4,10 +4,34 @@ declare(strict_types=1);
 
 namespace PhpLlm\LlmChain\ToolBox;
 
+use PhpLlm\LlmChain\ToolBox\Attribute\ToolParameter;
+
 /**
  * @phpstan-type ParameterDefinition array{
  *     type: 'object',
- *     properties: array<string, array{type: string, description: string}>,
+ *     properties: array<string, array{
+ *         type: string,
+ *         description: string,
+ *         enum?: list<string>,
+ *         const?: string|int|list<string>,
+ *         pattern?: string,
+ *         minLength?: int,
+ *         maxLength?: int,
+ *         minimum?: int,
+ *         maximum?: int,
+ *         multipleOf?: int,
+ *         exclusiveMinimum?: int,
+ *         exclusiveMaximum?: int,
+ *         minItems?: int,
+ *         maxItems?: int,
+ *         uniqueItems?: bool,
+ *         minContains?: int,
+ *         maxContains?: int,
+ *         required?: bool,
+ *         minProperties?: int,
+ *         maxProperties?: int,
+ *         dependentRequired?: bool,
+ *     }>,
  *     required: list<string>,
  * }
  */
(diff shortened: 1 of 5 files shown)


src/ToolBox/AsParameter.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpLlm\LlmChain\ToolBox;
```
OskarStark, line 5: Do you agree with the namespace @chr-hertel ?

chr-hertel: what do you think about introducing an `Attribute` namespace? too early? rather below `ToolBox` or on a higher level?

chr-hertel:
i guess i would prefer `PhpLlm\LlmChain\Attribute\ToolParameter` 🤔 

and i think it would be great to consider all the standardized cases:
image
see https://json-schema.org/draft/2019-09/json-schema-validation#rfc.section.6

src/ToolBox/AsParameter.php
OskarStark: ToolBox/Attribute suits me

OskarStark: Done

OskarStark:
> and i think it would be great to consider all the standardized cases:

In this case it would be good to a) map the type, so that `pattern` can only be used if it is a string or define the type, but that would double the noise.

src/ToolBox/ParameterAnalyzer.php
```diff
@@ -16,7 +18,7 @@ final class ParameterAnalyzer
     /**
      * @return ParameterDefinition|null
      */
-    public function getDefinition(string $className, string $methodName): ?array
+    private function getDefinition(string $className, string $methodName): ?array
```
OskarStark, line 21:
```suggestion
    public function getDefinition(string $className, string $methodName): ?array
```

src/ToolBox/ParameterAnalyzer.php
```diff
@@ -48,10 +50,39 @@ public function getDefinition(string $className, string $methodName): ?array
...
+                ] as $attributeName) {
+                    // check if property is set and not null
+                    if (!property_exists($attribute, $attributeName)) {
+                        throw new \InvalidArgumentException(sprintf('Attribute "%s" does not have property "%s"', ToolParameter::class, $attributeName));
+                    }
+
+                    if (null !== $attribute->$attributeName) {
+                        $property[$attributeName] = $attribute->$attributeName;
+                    }
+                }
```
chr-hertel, line 82: you could try
```suggestion
                $attributeState = array_filter((array) $attributes[0]->newInstance(), fn ($value) => null !== $value);
                $property = array_merge($property, $attributeState);
```

OskarStark: Works, thanks

tests/ToolBox/Attribute/ToolParameterTest.php
chr-hertel, line 1: i bet this one is generated :D

fair enough, but you could tell your little friend to use `#[Test]` instead of that `test*`-prefix

chr-hertel:
Nice one :muscle:

src/ToolBox/ParameterAnalyzer.php
```diff
@@ -4,10 +4,12 @@
 
 namespace PhpLlm\LlmChain\ToolBox;
 
+use PhpLlm\LlmChain\ToolBox\Attribute\ToolParameter;
+
 /**
  * @phpstan-type ParameterDefinition array{
  *     type: 'object',
- *     properties: array<string, array{type: string, description: string}>,
+ *     properties: array<string, array{type: string, description: string, pattern?: string}>,
```
OskarStark, line 12: todo

tests/ToolBox/Attribute/ToolParameterTest.php
OskarStark: It is, yes!

Done

OskarStark:
> Nice one 💪

Tough one

chr-hertel (approved):

PR #76, opened 2024-10-01
Change order of commands in `Makefile`

OskarStark:


OskarStark:
CleanShot 2024-10-01 at 11 56 43@2x

It would be helpful if you would remove the 1 Approval rule, thanks

chr-hertel (approved):

chr-hertel:
can you image adopting conventional commits?
https://www.conventionalcommits.org/en/v1.0.0/

adopting your commits is something i usually do with your MRs - maybe rather a spleen of mine, but i like to have it in the git history

PR #79, opened 2024-10-01
Filter out `Document`s without text in `DocumentEmbedder` + tests

OskarStark:


tests/DocumentEmbedderTest.php
```diff
@@ -0,0 +1,45 @@
...
+use PhpLlm\LlmChain\Document\Vector;
+use PhpLlm\LlmChain\DocumentEmbedder;
+use PhpLlm\LlmChain\EmbeddingsModel;
+use PhpLlm\LlmChain\Store\StoreInterface;
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\Attributes\Test;
+use PHPUnit\Framework\TestCase;
+
+#[CoversClass(DocumentEmbedder::class)]
+final class DocumentEmbedderTest extends TestCase
```
chr-hertel, line 17: with that level of mocking i would propose to consider it a unit test and add `#[Small]` attribute here

tests/DocumentEmbedderTest.php
```diff
@@ -0,0 +1,45 @@
...
+
+    protected function setUp(): void
+    {
+        $this->embeddings = $this->createMock(EmbeddingsModel::class);
+        $this->store = $this->createMock(StoreInterface::class);
+        $this->embedder = new DocumentEmbedder($this->embeddings, $this->store);
+    }
+
+    #[Test]
+    public function testEmbedSingleDocument(): void
```
chr-hertel, line 31: i guess gpt was not aware the `test*` prefix is not needed here? :D

OskarStark: done

tests/DocumentEmbedderTest.php
OskarStark: done, not really used to this attributes, sorry

examples/store-mongodb-similarity-search.php
```diff
@@ -55,7 +56,7 @@
 
 // create embeddings for documents
 $platform = new OpenAI(HttpClient::create(), $_ENV['OPENAI_API_KEY']);
-$embedder = new DocumentEmbedder($embeddings = new Embeddings($platform), $store);
+$embedder = new DocumentEmbedder($embeddings = new Embeddings($platform), $store, new NullLogger());
```
OskarStark, line 59: `NullLogger` is fine here?

tests/DocumentEmbedderTest.php
```diff
@@ -0,0 +1,93 @@
...
+
+    //    #[Test]
+    //    public function embedDocumentWithoutText(): void
+    //    {
+    //        $document = new Document(Uuid::v4(), null, null);
+    //        $this->embeddings->expects($this->once())->method('multiCreate');
+    //        $this->store->expects($this->once())->method('addDocuments');
+    //
+    //        $this->embedder->embed($document);
+    //    }
```
OskarStark, line 71: I don't think we want to can it if text is null, but that happens right now.

examples/store-mongodb-similarity-search.php
chr-hertel: could also be part of the constructor as default, but don't really have a preference tbh

tests/DocumentEmbedderTest.php
chr-hertel: log and skip?

OskarStark: that instead of an exception

OskarStark: done

examples/store-mongodb-similarity-search.php
OskarStark: will do

OskarStark: done

OskarStark:
Ready to merge from my side

chr-hertel (approved):

PR #85, opened 2024-10-01
Use dedicated Document class if vector is available

OskarStark:
Just a first idea, was this, what you had in mind?

### DIFF
--- a/src/Document/VectorizedDocument.php
+++ b/src/Document/VectorizedDocument.php
@@ -1,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpLlm\LlmChain\Document;
+
+use Symfony\Component\Uid\Uuid;
+use Symfony\Component\Uid\UuidV4;
+
+final class VectorizedDocument extends Document
+{
+    public function __construct(
+        public Uuid $id,
+        public ?string $text,
+        public ?Vector $vector,
+        public Metadata $metadata = new Metadata([]),
+    ) {
+        parent::__construct($id, $text, $metadata);
+    }
+
+    public function hasVector(): bool
+    {
+        return $this->vector instanceof Vector;
+    }
+}
--- a/src/Document/Document.php
+++ b/src/Document/Document.php
@@ -7,44 +7,13 @@ namespace PhpLlm\LlmChain\Document;
 use Symfony\Component\Uid\Uuid;
 use Symfony\Component\Uid\UuidV4;
 
-final readonly class Document
+class Document
 {
     public function __construct(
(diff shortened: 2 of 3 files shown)


chr-hertel:
Yes, i think inheritance, also maybe a solution for `Message` is a valid solution here

src/Document/VectorizedDocument.php
```diff
@@ -0,0 +1,25 @@
...
+
+use Symfony\Component\Uid\Uuid;
+use Symfony\Component\Uid\UuidV4;
+
+final class VectorizedDocument extends Document
+{
+    public function __construct(
+        public Uuid $id,
+        public ?string $text,
+        public ?Vector $vector,
```
chr-hertel, line 15: having this nullable feels wrong with the name. and the `hasVector` could be replaced in user land code with `$document instanceof VectorizedDocument`

chr-hertel:
I would go even one step further and change the `StoreInterface` to rely on the `VectorizedDocument`

chr-hertel:
or is it an `EmbeddedDocument`? :thinking:

src/Document/VectorizedDocument.php
OskarStark: Indeed 👍

OskarStark:
> or is it an EmbeddedDocument? 🤔

Yes, why not, makes sense.

Another thing is, how we want to deal with a text? it is nullable, so what's the purpose here? should we make it required as a string in `Document`? I mean we need anyway sth. which we can use for embedding. Maybe renaming it to `$content` makes more sense?

chr-hertel:
so the `text` is the input for the embedding. and in my understanding mostly created by (a subset) of data points of an entity or the metadata.

using your example:
```php
class Movie
{
    public string $title;
    public string $director;
    public string $description;
}
```

in simple cases we could just convert it into a text to embed like this
```
Title: {$movie->title}
Director: {$movie->director}
Description:  {$movie->description}
```
but that could a) vary on the state or meaning of the base object or b) involve other strategies like enhancing the content dynamically

OskarStark:
Yes, I would decouple the metadata from the used string for generating the embedding.

That's why I propose to rename text to content.

In the metadata one could store what he wants, but the content string should be produced in userland without any magic IMHO

OskarStark:
This PR is replaced by 
* https://github.com/php-llm/llm-chain/pull/95

chr-hertel:
so we can close this and #95 and #101 belong together?

the comments are confusing me :D

OskarStark:
Yes this was the first idea here to tackle this topic.

Issue #77, opened 2024-10-01
Introduce UID for Messages


OskarStark:
We propose introducing UIDs for messages, which would be particularly beneficial in a frontend context. This approach would enable more efficient message handling by reducing unnecessary back-and-forth communication between the client and backend, especially when only the latest updates need to be tracked or pushed to the frontend.

Whenever responding to the frontend, we could push only the new messages.

The UID should ideally be generated based on the content of the Message. Initially, I considered using ULIDs, as they are sortable. However, ULIDs only work when generated on the fly with new, which isn’t feasible for our case since we need to be able to “recalculate” the UID when necessary.

Otherwise, we’d need to implement our own logic to handle this.

cc @silasjoisten as you’re working on the frontend
@DZunke for your awareness

One thing we’re uncertain about is whether we can send UIDs to the LLM or if they should be removed before querying the LLM.

DZunke:
Would it really be useful to implement UID within this library? Just asking because the library is not delivering a persistence, so every message in the bag would get a UID, yes, but for what would it be utilized? 

I think the UID is more useful when you bring this stateless library to a point to have a state. This is why i have wrapped an `ExtendedMessage` around the Messages that enhances the Message, that is put in there, with a UID to make the message storable in the storage of my application and so to have always the same identifier for each message. 

Also when you want to just push new messages to a frontend you would have to have a storage with the already sent messages, or at least, identifiers in it, but, as mentioned ... such persistence is from my PoV more a concern of the application implementing this library 🤔

silasjoisten:
Having a UID natively generated by the library can significantly simplify and standardize how messages are managed across various applications using the library. While it’s true that applications can wrap messages and generate their own UIDs, having a native solution reduces the burden on developers to implement this logic repeatedly in every use case. It ensures consistency in how messages are identified and tracked, regardless of the implementation context.

Moreover, even if the library is stateless, the UIDs help in creating a more predictable and reliable flow of messages between different parts of an application (or even between different systems). This is especially important when dealing with distributed or real-time systems, where keeping track of the order and uniqueness of messages is critical. In these cases, the frontend can easily recognize and handle only new messages or changes without additional overhead in comparing message content manually.

Additionally, having a UID allows for the potential future enhancement of the library, like built-in persistence layers, caching strategies, or easier debugging capabilities, since each message could be traced back more easily. Overall, the idea is to provide a more cohesive and feature-complete experience to developers, who can then build upon it without reinventing the wheel each time.

[...]

PR #92, opened 2024-10-01
feat: introduce message metadata and traits (#91)

chr-hertel:
Multiple ideas around Message - started by #77 

1. i think the traits make it easier for people to build their own messages on top of the interface if needed
2. the metadata are a low structured gimmick that could reduce the need of having custom messages

didn't that anything yet and just a draft ...


OskarStark:
TBH I am not sure I like this 😄

chr-hertel:
Me neither. What is your concern about?

chr-hertel:
Imagine i would want that UID in all messages, i would need to extend all messages to do so. Custom Systemmessage, CustomAssistantMessage, etc... or when using metadata the data would rather be unstructured

chr-hertel:
And that wouldn't be sufficient since processors, like the Tool one, add built in messages without UID.

But other than UID it could be the actual User reference, additional information about the tool call, like used source documents or links

chr-hertel:
closed in favor of #166 and #174

PR #94, opened 2024-10-02
Clear exception hierachy

OskarStark:
We should rely on our own exception classes


OskarStark:
Ready to merge from my side

chr-hertel (approved):

PR #95, opened 2024-10-02
[BC Break] Require `Document::$text` contains non empty string and simplify `Document` API

OskarStark:
I would do this as a first step before continuing in
* #85 

In the next step I would remove the `vector` property, so we always have a `Document` in a valid state and then create `EmbeddedDocument` class which extends `Document` and add the `vector` information.

With this change, a `Document` is always embeddable: Right now, this is not the case!

### DIFF
--- a/src/Store/Azure/SearchStore.php
+++ b/src/Store/Azure/SearchStore.php
@@ -78,6 +78,7 @@ final readonly class SearchStore implements VectorStoreInterface
         return array_merge([
             'id' => $document->id,
             $this->vectorFieldName => $document->vector->getData(),
+            'text' => $document->text,
         ], $document->metadata->getArrayCopy());
     }
 
@@ -88,8 +89,8 @@ final readonly class SearchStore implements VectorStoreInterface
     {
         return new Document(
             id: Uuid::fromString($data['id']),
-            text: null,
-            vector: null,
+            text: $data['text'],
+            vector: $data[$this->vectorFieldName] ? new Vector($data[$this->vectorFieldName]) : null,
             metadata: new Metadata($data),
         );
     }
--- a/src/Store/ChromaDB/Store.php
+++ b/src/Store/ChromaDB/Store.php
@@ -55,10 +55,11 @@ final readonly class Store implements VectorStoreInterface
 
         $documents = [];
         for ($i = 0; $i < count($queryResponse->metadatas[0]); ++$i) {
-            $documents[] = Document::fromVector(
-                new Vector($queryResponse->embeddings[0][$i]),
-                Uuid::fromString($queryResponse->ids[0][$i]),
-                new Metadata($queryResponse->metadatas[0][$i]),
+            $documents[] = new Document( // not sure how to insert and get the text here @chr-hertel
+                id: Uuid::fromString($queryResponse->ids[0][$i]),
+                text: '???',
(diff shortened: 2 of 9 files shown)


src/Store/ChromaDB/Store.php
```diff
@@ -55,10 +55,11 @@ public function query(Vector $vector, array $options = []): array
 
         $documents = [];
         for ($i = 0; $i < count($queryResponse->metadatas[0]); ++$i) {
-            $documents[] = Document::fromVector(
-                new Vector($queryResponse->embeddings[0][$i]),
-                Uuid::fromString($queryResponse->ids[0][$i]),
-                new Metadata($queryResponse->metadatas[0][$i]),
+            $documents[] = new Document( // not sure how to insert and get the text here @chr-hertel
+                id: Uuid::fromString($queryResponse->ids[0][$i]),
+                text: '???',
```
OskarStark, line 60: - [ ] How could we add the text to Chroma?

cc @chr-hertel

src/Store/ChromaDB/Store.php
```diff
@@ -55,10 +55,11 @@ public function query(Vector $vector, array $options = []): array
 
         $documents = [];
         for ($i = 0; $i < count($queryResponse->metadatas[0]); ++$i) {
-            $documents[] = Document::fromVector(
-                new Vector($queryResponse->embeddings[0][$i]),
-                Uuid::fromString($queryResponse->ids[0][$i]),
-                new Metadata($queryResponse->metadatas[0][$i]),
+            $documents[] = new Document( // not sure how to insert and get the text here @chr-hertel
```
OskarStark, line 58:
```suggestion
            $documents[] = new Document(
```
?

src/Store/Pinecone/Store.php
```diff
@@ -63,10 +64,11 @@ public function query(Vector $vector, array $options = []): array
...
         foreach ($response->json()['matches'] as $match) {
-            $documents[] = Document::fromVector(
-                new Vector($match['values']),
-                Uuid::fromString($match['id']),
-                new Metadata($match['metadata']),
+            $documents[] = new Document(
+                id: Uuid::fromString($match['id']),
+                text: $match['text'],
+                vector: new Vector($match['values']),
+                metadata: new Metadata($match['metadata']),
```
OskarStark, line 71: Does this work @chr-hertel ?

src/Store/ChromaDB/Store.php
chr-hertel: i guess we would store the text in chroma. not doing it atm at least.

src/Store/Pinecone/Store.php
chr-hertel: nope

src/Store/Azure/SearchStore.php
```diff
@@ -88,8 +89,8 @@ private function convertArrayToDocument(array $data): Document
     {
         return new Document(
             id: Uuid::fromString($data['id']),
-            text: null,
-            vector: null,
+            text: $data['text'],
+            vector: $data[$this->vectorFieldName] ? new Vector($data[$this->vectorFieldName]) : null,
```
OskarStark, line 93: hmm I would not go this direction here, I would assume, that `$data[$this->vectorFieldName]` is always available.

OskarStark:
Superseded by
* #110

PR #100, opened 2024-10-02
Simplify `SchemaFactory`

OskarStark:
Or is there a reason for the factory method @chr-hertel ?


chr-hertel (approved):
no particular reason :+1:

PR #101, opened 2024-10-02
Introduce `EmbeddedDocument`

OskarStark:
Can be merged after
* #85

### DIFF
--- a/src/Document/EmbeddedDocument.php
+++ b/src/Document/EmbeddedDocument.php
@@ -1,0 +1,34 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpLlm\LlmChain\Document;
+
+use Symfony\Component\Uid\Uuid;
+
+final readonly class EmbeddedDocument extends Document
+{
+    public function __construct(
+        Uuid $id,
+        string $text,
+        public Vector $vector,
+        Metadata $metadata = new Metadata([]),
+    ) {
+        parent::__construct($id, $text, $metadata);
+    }
+
+    public function withVector(Vector $vector): self
+    {
+        return new self(
+            $this->id,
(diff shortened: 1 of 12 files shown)


src/Document/EmbeddedDocument.php
```diff
@@ -0,0 +1,34 @@
...
+            $this->text,
+            $vector,
+            $this->metadata,
+        );
+    }
+
+    public function hasVector(): bool
+    {
+        return null !== $this->vector;
+    }
```
OskarStark, line 33:
```suggestion

```

chr-hertel:
what do you think about like a tupel of document classes:

`TextDocument` and `VectorDocument` - both have an UUID and Metadata

```php
/**
 * @param TextDocument|TextDocument[] $documents
 */
DocumentEmbedder::create(TextDocument|array $documents, ...): void

// only converting than

$vectorDocuments[] = new VectorDocument($vectors[$i], $document->id, $document->metadata);

// and drop text in store context completely

/**
 * @param VectorDocument|VectorDocument[] $documents
 **/
StoreInterface::add(VectorDocument|array $documents): void;

/**
 * @return VectorDocument[]
 */
VectorStoreInterface::query(Vector $vector, array $options = []): array;
```

OskarStark:
I like it, I thought sth like this without inheritance too in the beginning

OskarStark:
No I don't think we need that abstraction upstream

chr-hertel:
you want to tackle that with your PRs or shall I?

OskarStark:
If you have time, go ahead. Either use my PR or open a new one where you do the transformation completely

OskarStark:
Superseded by
* #110

PR #104, opened 2024-10-02
Rename `ContentInterface` to `Content`

OskarStark:
Extracted from
* https://github.com/php-llm/llm-chain/pull/92

I like it !

### DIFF
--- a/src/Message/UserMessage.php
+++ b/src/Message/UserMessage.php
@@ -4,19 +4,19 @@ declare(strict_types=1);
 
 namespace PhpLlm\LlmChain\Message;
 
-use PhpLlm\LlmChain\Message\Content\ContentInterface;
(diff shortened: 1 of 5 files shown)


src/Message/UserMessage.php
```diff
@@ -4,19 +4,19 @@
...
 {
     /**
-     * @var list<ContentInterface>
+     * @var list<Content>
      */
     public array $content;
 
     public function __construct(
-        ContentInterface ...$content,
+        Content ...$content,
```
OskarStark, line 19: I will add the `string` and `\Stringable` type in a follow-up PR

OskarStark: * https://github.com/php-llm/llm-chain/issues/105

chr-hertel (approved):

PR #112, opened 2024-10-03
Use `#[\SensitiveParameter]`

OskarStark:


src/OpenAI/Platform/Azure.php
```diff
@@ -18,7 +18,7 @@ public function __construct(
         private string $baseUrl,
         private string $deployment,
         private string $apiVersion,
-        private string $key,
+        #[\SensitiveParameter] private string $apiKey,
```
OskarStark, line 21: Needs to be adjusted in bundle config

chr-hertel: after the next tag the bundle needs some love anyways :+1:

OskarStark: Prepared in
* https://github.com/php-llm/llm-chain-bundle/pull/38

chr-hertel (approved):
good catch :+1:

src/OpenAI/Platform/Azure.php
OskarStark: > after the next tag the bundle needs some love anyways 👍

Yes, lets do it in
* https://github.com/php-llm/llm-chain-bundle/pull/38

PR #113, opened 2024-10-04
Fix assume vector key is available

OskarStark:
If it is not available, `null` is an invalid type for our `vector property`

CleanShot 2024-10-04 at 10 27 38@2x

### DIFF
--- a/src/Store/Azure/SearchStore.php
+++ b/src/Store/Azure/SearchStore.php
@@ -80,7 +80,7 @@ final readonly class SearchStore implements VectorStoreInterface
     {
         return new VectorDocument(
             id: Uuid::fromString($data['id']),
-            vector: $data[$this->vectorFieldName] ? new Vector($data[$this->vectorFieldName]) : null,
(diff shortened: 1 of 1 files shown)


src/Store/Azure/SearchStore.php
```diff
@@ -78,9 +79,13 @@ private function convertToIndexableArray(VectorDocument $document): array
      */
     private function convertToVectorDocument(array $data): VectorDocument
     {
+        if (!array_key_exists($this->vectorFieldName, $data)) {
```
OskarStark, line 82: A little verbose, but it should be helpful IMHO

chr-hertel:
what do you think about having a `NullVector` as a fallback throwing that exception? a bit more generic than, but delaying the issue to the point users really want to interact with the vector data.

not sure we need to enforce the vector being around after querying - don't have a use case on my mind at the moment

OskarStark:
Superseded by
* #114

PR #114, opened 2024-10-04
Introduce `NullVector` and `VectorInterface`

OskarStark:
If it is not available, `null` is an invalid type for our `vector property`

CleanShot 2024-10-04 at 10 27 38@2x

### DIFF
--- a/src/Document/NullVector.php
+++ b/src/Document/NullVector.php
@@ -4,15 +4,16 @@ declare(strict_types=1);
 
 namespace PhpLlm\LlmChain\Document;
 
+use PhpLlm\LlmChain\Exception\RuntimeException;
+
 final class NullVector implements VectorInterface
 {
     public function getData(): array
     {
-        return [];
+        throw new RuntimeException('getData() method cannot be called on a NullVector.');
     }
 
     public function getDimensions(): int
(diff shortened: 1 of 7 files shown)


OskarStark:
Replaces 
* #113

src/Document/NullVector.php
```diff
@@ -0,0 +1,18 @@
...
+
+declare(strict_types=1);
+
+namespace PhpLlm\LlmChain\Document;
+
+final class NullVector implements VectorInterface
+{
+    public function getData(): array
+    {
+        return [];
```
OskarStark, line 11: Or better throw here?

src/Document/NullVector.php
```diff
@@ -0,0 +1,18 @@
...
+    }
+
+    public function getDimensions(): int
+    {
+        return 0;
```
OskarStark, line 16: and here?

src/Document/NullVector.php
chr-hertel: both fine i'd say

OskarStark:
Ready to merge from my side

chr-hertel (approved):

PR #116, opened 2024-10-04
chore: remove reasoning example to safe my credit card

chr-hertel:
this is just crazy expensive compared to the other examples ...

image


OskarStark:
Let's keeping but let's add another env var which enables the run of this example.

To me it is great to have this example ✌🏻

PR #117, opened 2024-10-05
refactor: `ResponseInterface` as `Chain` return value

chr-hertel:


OskarStark:
I like this

OskarStark:
I added some tests for your branch here:
* https://github.com/php-llm/llm-chain/pull/120

OskarStark (approved):

OskarStark:
I think we can add `No BC break` label instead of adding `BC break` label 🤣

chr-hertel:
i'm not awake enough for this :D why is it not a BC break?

OskarStark:
I mean we nearly only introduce BC breaks, so maybe its better to opt out 😅

chr-hertel:
true :angel:

PR #119, opened 2024-10-05
Revert "chore: remove reasoning example to safe my credit card (#116)"

OskarStark:
This reverts commit 4a48bc4f875f9de1aaa9e1d1016980b677eec7b1 and added checks via ENV var

Feel free to close

### DIFF
--- a/.env
+++ b/.env
@@ -24,3 +24,6 @@ MONGODB_URI=
 # For using Pinecone (store)
 PINECONE_API_KEY=
 PINECONE_HOST=
+
+# Some examples are expensive to run, so we disable them by default
+RUN_EXPENSIVE_EXAMPLES=false
--- a/examples/structured-output-math.php
+++ b/examples/structured-output-math.php
@@ -23,7 +23,7 @@ if (empty($_ENV['OPENAI_API_KEY'])) {
     exit(1);
 }
 
-if (empty($_ENV['RUN_EXPENSIVE_EXAMPLES']) || filter_var($_ENV['RUN_EXPENSIVE_EXAMPLES'], FILTER_VALIDATE_BOOLEAN) === false) {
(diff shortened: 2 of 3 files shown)


.env
```diff
@@ -24,3 +24,6 @@ MONGODB_URI=
 # For using Pinecone (store)
 PINECONE_API_KEY=
 PINECONE_HOST=
+
+# Some examples are expensive to run, so we disable them by default
+RUN_EXPENSIVE_EXAMPLES=false
```
chr-hertel, line 29: :D love it

examples/structured-output-math.php
```diff
@@ -23,6 +23,11 @@
     exit(1);
 }
 
+if (empty($_ENV['RUN_EXPENSIVE_EXAMPLES']) || false === filter_var($_ENV['RUN_EXPENSIVE_EXAMPLES'], FILTER_VALIDATE_BOOLEAN)) {
+    echo 'This example is marked as expensive and will not run unless RUN_EXPENSIVE_EXAMPLES is set to true.'.PHP_EOL;
+    exit(1);
+}
+
```
chr-hertel, line 30: wrong example

.env
chr-hertel: maybe rather an option on the example runner?

examples/structured-output-math.php
OskarStark: indeed

chr-hertel (approved):

.env
OskarStark: I think its safe here, at least I, run the example one by one via PHP instead of the runner

OskarStark: > :D love it

I think those examples are key and the best documentation for the users

PR #122, opened 2024-10-05
feat: support creation of image from binary file

chr-hertel:


src/Message/Content/Image.php
```diff
@@ -14,6 +16,17 @@ public function __construct(
     ) {
     }
 
+    public static function fromFile(string $filePath): self
```
OskarStark, line 19: I would rather check in the ctor and do the logic if the string does not start with http or data and then do the logic. Another idea is to check if it starts with a slash

OskarStark (approved):
Besides my comment, good to merge 🙏

PR #123, opened 2024-10-05
refactor: use event dispatcher instead of processor interfaces

chr-hertel:


chr-hertel:
having a hard time from an architecture point sacrifizing explicit extension points over that more generic event approach - and on the other hand it's so common in symfony ecosystem.

chr-hertel:
let's postpone this discussion until we know more about platform and model abstractions. i guess events can be very handy there as well.

chr-hertel:
closed in favor of #140

PR #129, opened 2024-10-05
LLama prompt

OskarStark:
Untested, but this could do the trick

### DIFF
--- a/src/Model/Language/Llama.php
+++ b/src/Model/Language/Llama.php
@@ -5,7 +5,11 @@ declare(strict_types=1);
 namespace PhpLlm\LlmChain\Model\Language;
 
 use PhpLlm\LlmChain\LanguageModel;
+use PhpLlm\LlmChain\Message\AssistantMessage;
+use PhpLlm\LlmChain\Message\Message;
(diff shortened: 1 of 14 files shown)


src/Model/Language/Llama.php
```diff
@@ -22,14 +26,37 @@ public function call(MessageBag $messages, array $options = []): TextResponse
...
     }
 
+    private static function convertToPrompt(MessageBag $messageBag): string
+    {
+        $messages = [];
+
+        /** @var MessageInterface $message */
+        foreach ($messageBag->getIterator() as $message) {
+            if ($message instanceof UserMessage) {
+                $content = $message->content[0]->text;
```
OskarStark, line 45: I think this is not enough here, because of the N-Contents we could have

chr-hertel:
i'm afraid that's gonna be more complex:
https://www.llama.com/docs/model-cards-and-prompt-formats/llama3_1/#-supported-roles-

OskarStark:
Ah I remember, I mixed sth. up, sorry.

OskarStark:
Replaced by 
* #135

PR #136, opened 2024-10-07
Fix test class name

OskarStark:


chr-hertel:
Good catch, but file name is still off

OskarStark:
What do you mean?

chr-hertel (approved):

PR #140, opened 2024-10-13
refactor: introduce decoupled platform layer and convert models to state objects

chr-hertel:
this PR focusses on shaping the model and platform concept.

* models are only state objects now and the service character got killed in favor of Platform extension points `ResponseFactory` and `ResponseConverter` => decoupling => way better testability
* opens up the platform requests to be quite generic => leeway to supporting large platforms like ollama, huggingface or  replicate in a generic manner
* bridge namespace heavily inspired by Symfony

Only thing missing right now is fixing embedding by bringing support for parallel calls on Platform back

_best to review in IDE_


src/Bridge/Azure/OpenAI/PlatformFactory.php
```diff
@@ -0,0 +1,29 @@
...
+        #[\SensitiveParameter] string $apiKey,
+    ): Platform {
+        $httpClient = new EventSourceHttpClient();
+        $embeddingsResponseFactory = new EmbeddingsResponseFactory($httpClient, $baseUrl, $deployment, $apiVersion, $apiKey);
+        $GPTResponseFactory = new GPTResponseFactory($httpClient, $baseUrl, $deployment, $apiVersion, $apiKey);
+
+        return new Platform(
+            [$GPTResponseFactory, $embeddingsResponseFactory],
+            [new GPT\ResponseConverter(), new Embeddings\ResponseConverter()],
+        );
```
chr-hertel, line 27: one important benefit:
response factory for the Azure specific platform is different, but response factory can sit on the openai default one

OskarStark:
Sorry for being not so active, I am a little bit busy right now.
I didn't find time to check this PR in detail

chr-hertel:
@OskarStark no worries at all Oskar - i'm currently finishing up those 4 MRs and will try them in my demo app with the bundle to get them merged :+1:

OskarStark:
Thanks, if my response format is working for tools, I could also switch entirely to our bundle/lib

PR #142, opened 2024-11-12
Fix empty tool list in request is ending in 400 bad request

DZunke:
With the empty tools array in the request to Open AI the request is failing as a bad request. When the tool box is empty it should be ignored, even the processor is registered to the chain.

### DIFF
--- a/src/ToolBox/ChainProcessor.php
+++ b/src/ToolBox/ChainProcessor.php
@@ -29,8 +29,13 @@ final class ChainProcessor implements InputProcessor, OutputProcessor, ChainAwar
             throw MissingModelSupport::forToolCalling($input->llm::class);
         }
 
+        $toolMap = $this->toolBox->getMap();
+        if ([] === $toolMap) {
+            return;
+        }
+
         $options = $input->getOptions();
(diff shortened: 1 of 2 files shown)


DZunke:
Haha. @chr-hertel  haven't you mentioned there is nothing happening with this automatic review tool above? I think it works now with a wall of text. 🤣 

And for the failing pipeline see https://github.com/php-llm/llm-chain/pull/143

chr-hertel:
Wow, not sure it is helpful tho :joy: 

so, content-wise i'll give it rather only half point - that bot thing is not happy with you not checking the validity of the map further - and i'd say i get where this is coming from, but would rather not check the tool map value at all but throw an exception in the ToolBox and catch that --- buuut, that change would be larger :D

in general it is way too much text as that i would appreciate that bot review for now. just tooo much boilerplate.

chr-hertel (approved):
needs a rebase

DZunke:
Yeah. The Review itself is not that bad because all is justified and not that bad explanations but too much text for such a simple change. I thik it could be very hard for a large PR.

And i have rebased 😊

chr-hertel:
Thanks @DZunke

PR #143, opened 2024-11-12
Conflict with PHPDocumentor higher then 5.5 because of doc parser bugs

DZunke:
There seem to be an error in the phpdocumentor of version 5.6 upwards which utilizes PHPStan phpdoc-parser in the fresh released version 2 and there is a compability problem. With 5.5 which does not allow the phpdoc-parser in v2 it is working as expected again. 

May need some time to do all the internal fixes before update.

See: 

```
There was 1 failure:

1) PhpLlm\LlmChain\Tests\StructuredOutput\ChainProcessorTest::processInputThrowsExceptionWhenLlmDoesNotSupportStructuredOutput
Failed asserting that exception of type "ArgumentCountError" matches expected exception "PhpLlm\LlmChain\Exception\MissingModelSupport". Message was: "Too few arguments to function PHPStan\PhpDocParser\Lexer\Lexer::__construct(), 0 passed in /home/runner/work/llm-chain/llm-chain/vendor/symfony/type-info/TypeResolver/StringTypeResolver.php on line 66 and exactly 1 expected" at
/home/runner/work/llm-chain/llm-chain/vendor/phpstan/phpdoc-parser/src/Lexer/Lexer.php:102
/home/runner/work/llm-chain/llm-chain/vendor/symfony/type-info/TypeResolver/StringTypeResolver.php:66
/home/runner/work/llm-chain/llm-chain/vendor/symfony/type-info/TypeResolver/TypeResolver.php:71
/home/runner/work/llm-chain/llm-chain/vendor/symfony/type-info/TypeResolver/TypeResolver.php:66
/home/runner/work/llm-chain/llm-chain/vendor/symfony/property-info/Extractor/ReflectionExtractor.php:102
/home/runner/work/llm-chain/llm-chain/vendor/symfony/property-access/PropertyAccessor.php:91
/home/runner/work/llm-chain/llm-chain/vendor/symfony/property-access/PropertyAccessorBuilder.php:289
/home/runner/work/llm-chain/llm-chain/vendor/symfony/property-access/PropertyAccess.php:26
/home/runner/work/llm-chain/llm-chain/vendor/symfony/serializer/Normalizer/ObjectNormalizer.php:53
/home/runner/work/llm-chain/llm-chain/tests/StructuredOutput/ChainProcessorTest.php:73
```

PR #144, opened 2024-11-12
Add interface to final readonly chain for testability

DZunke:
For simple mocks it would be cool to have the interface if the class is needed to be final and readonly as it is not a DTO but a service 😉 

Tests are failing cause of https://github.com/php-llm/llm-chain/pull/143


chr-hertel (approved):
valid, thanks!

PR #150, opened 2024-12-05
Parse OpenAI response embedding to contain all embeddings

DZunke:
When i request the embeddings with mutliple texts the response only contaion the first entry of the response. I did not really understand why it should be this way as the `VectorResponse` already supports having multiple vectors and so have adapted the response converter for the OpenAi Bridge embeddings. 

```php
$texts = array_fill(0, 20, 'Foo Bar Baz');
$vectors = $platform->request(new Embeddings(), $texts)->getContent();
```

The response of such a request, utilizing the OpenAI embeddings is an array with a single text or an array of text. So it does not change the current result of the `VectorResponse` for the user of the request.

Hope this fits?

### DIFF
--- a/tests/Bridge/OpenAI/Embeddings/ResponseConverterTest.php
+++ b/tests/Bridge/OpenAI/Embeddings/ResponseConverterTest.php
@@ -1,0 +1,56 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpLlm\LlmChain\Tests\Bridge\OpenAI\Embeddings;
+
+use PhpLlm\LlmChain\Bridge\OpenAI\Embeddings\ResponseConverter;
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\Attributes\Small;
+use PHPUnit\Framework\Attributes\Test;
+use PHPUnit\Framework\TestCase;
+use Symfony\Contracts\HttpClient\ResponseInterface;
+
+#[CoversClass(ResponseConverter::class)]
+#[Small]
+class ResponseConverterTest extends TestCase
+{
+    #[Test]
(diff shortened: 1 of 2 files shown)


tests/Bridge/OpenAI/Embeddings/ResponseConverterTest.php
```diff
@@ -0,0 +1,56 @@
...
+use Symfony\Contracts\HttpClient\ResponseInterface;
+
+#[CoversClass(ResponseConverter::class)]
+#[Small]
+class ResponseConverterTest extends TestCase
+{
+    #[Test]
+    public function itConvertsAResponseToAVectorResponse(): void
+    {
+        $response = $this->createStub(ResponseInterface::class);
```
OskarStark, line 21: Why not using a JsonMockResponse directly?

DZunke: Do you maybe have more information to get me on your level here? I am a bit stuck with that you could mean with a `JsonMockResponse` 🤔

OskarStark: CleanShot 2024-12-06 at 12 49 02@2x

chr-hertel: i think in this case the benefit is not really there yet - more like a question of style, e.g. not using mock frameworks but implementations.

in general i can think of some infrastructure for MockResponse from file where also processing is done already - when oop design is done, it could be a quite common scenario to test those response converters and contract is `ResponseInterface` yup

chr-hertel (approved):
Yes, i guess that was just a short coming - thanks for the patch & test!

PR #158, opened 2024-12-16
Handle response errors

Iyadhfaleh:
I think json_decode is better to handle responses APIs . 
"toArray" method will throw exception when status code >= 300 .

### DIFF
--- a/src/Bridge/Anthropic/ModelHandler.php
+++ b/src/Bridge/Anthropic/ModelHandler.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace PhpLlm\LlmChain\Bridge\Anthropic;
 
+use http\Exception\RuntimeException;
 use PhpLlm\LlmChain\Model\Message\MessageBag;
 use PhpLlm\LlmChain\Model\Model;
 use PhpLlm\LlmChain\Model\Response\ResponseInterface as LlmResponse;
@@ -14,6 +15,10 @@ use PhpLlm\LlmChain\Platform\ResponseConverter;
 use Symfony\Component\HttpClient\Chunk\ServerSentEvent;
 use Symfony\Component\HttpClient\EventSourceHttpClient;
 use Symfony\Component\HttpClient\Exception\JsonException;
+use Symfony\Contracts\HttpClient\Exception\ClientExceptionInterface;
+use Symfony\Contracts\HttpClient\Exception\RedirectionExceptionInterface;
+use Symfony\Contracts\HttpClient\Exception\ServerExceptionInterface;
+use Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface;
 use Symfony\Contracts\HttpClient\HttpClientInterface;
 use Symfony\Contracts\HttpClient\ResponseInterface;
 use Webmozart\Assert\Assert;
@@ -55,13 +60,27 @@ final readonly class ModelHandler implements ModelClient, ResponseConverter
         ]);
     }
 
+    /**
+     * @throws TransportExceptionInterface
+     * @throws ServerExceptionInterface
+     * @throws RedirectionExceptionInterface
+     * @throws ClientExceptionInterface
+     */
     public function convert(ResponseInterface $response, array $options = []): LlmResponse
     {
         if ($options['stream'] ?? false) {
             return new StreamResponse($this->convertStream($response));
         }
 
-        $data = $response->toArray();
+        $data = json_decode($response->getContent(false), true);
+
+        if (!isset($data['content']) || !count($data['content'])) {
+            throw new RuntimeException('Response does not contain any content');
+        }
+
+        if (!isset($data['content'][0]['text'])) {
+            throw new RuntimeException('Response content does not contain any text');
+        }
 
         return new TextResponse($data['content'][0]['text']);
     }
--- a/src/Bridge/Anthropic/ModelHandler.php
+++ b/src/Bridge/Anthropic/ModelHandler.php
@@ -72,7 +72,7 @@ final readonly class ModelHandler implements ModelClient, ResponseConverter
             return new StreamResponse($this->convertStream($response));
         }
 
-        $data = json_decode($response->getContent(false), true);
+        $data = $response->toArray(false);
 
         if (!isset($data['content']) || !count($data['content'])) {
             throw new RuntimeException('Response does not contain any content');
--- a/src/Bridge/Anthropic/ModelHandler.php
+++ b/src/Bridge/Anthropic/ModelHandler.php
@@ -4,7 +4,7 @@ declare(strict_types=1);
 
 namespace PhpLlm\LlmChain\Bridge\Anthropic;
 
-use http\Exception\RuntimeException;
+use PhpLlm\LlmChain\Exception\RuntimeException;
 use PhpLlm\LlmChain\Model\Message\MessageBag;
 use PhpLlm\LlmChain\Model\Model;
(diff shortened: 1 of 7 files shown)


src/Bridge/Anthropic/ModelHandler.php
```diff
@@ -55,13 +60,27 @@ public function request(Model $model, object|array|string $input, array $options
...
+     * @throws RedirectionExceptionInterface
+     * @throws ClientExceptionInterface
+     */
     public function convert(ResponseInterface $response, array $options = []): LlmResponse
     {
         if ($options['stream'] ?? false) {
             return new StreamResponse($this->convertStream($response));
         }
 
-        $data = $response->toArray();
```
OskarStark, line 64: What about `toArray(throw: false)` ?

Iyadhfaleh: You are right , done

chr-hertel:
So, first of all thanks and I do agree that error handling is currently not in great shape.

the code that you added is mostly sitting on top of the expected json/array structure - which is a valid addition, but i doesn't help with 4xx/5xx situations, since everything is converted to a runtime exception and context of the original error or unexpected behavior would get lost.

i consider the structural changes that you check and error handling with RuntimeException a valuable addition, but would still keep the `toArray()` with throwing exceptions. nevertheless, this should be improved but rather in a different layer. and without loosing the original exception information.

this was the original handling, which is broken since i introduced the `AsyncResponse` i guess...
https://github.com/php-llm/llm-chain/blob/main/src/Platform.php#L44-L54

src/Bridge/Anthropic/ModelHandler.php
```diff
@@ -4,6 +4,7 @@
 
 namespace PhpLlm\LlmChain\Bridge\Anthropic;
 
+use http\Exception\RuntimeException;
```
chr-hertel, line 7: use `PhpLlm\LlmChain\Exception\RuntimeException` instead

src/Bridge/Anthropic/ModelHandler.php
```diff
@@ -55,13 +61,28 @@ public function request(Model $model, object|array|string $input, array $options
         ]);
     }
 
+    /**
+     * @throws TransportExceptionInterface
+     * @throws ServerExceptionInterface
+     * @throws RedirectionExceptionInterface
+     * @throws DecodingExceptionInterface
+     * @throws ClientExceptionInterface
+     */
```
chr-hertel, line 70: please remove this again. this is only needed due to PhpStorm inspections, not because of adding value. if you want to introduce valuable error information

src/Bridge/Anthropic/ModelHandler.php
```diff
@@ -55,13 +61,28 @@ public function request(Model $model, object|array|string $input, array $options
...
     public function convert(ResponseInterface $response, array $options = []): LlmResponse
     {
         if ($options['stream'] ?? false) {
             return new StreamResponse($this->convertStream($response));
         }
 
-        $data = $response->toArray();
+        $data = $response->toArray(false);
```
chr-hertel, line 77: let's revert this change but keep the structural checks below

```suggestion
        $data = $response->toArray();
```

src/Bridge/Anthropic/ModelHandler.php
```diff
@@ -55,13 +61,28 @@ public function request(Model $model, object|array|string $input, array $options
...
+
+        if (!isset($data['content']) || !count($data['content'])) {
```
chr-hertel, line 79: not a big fan of implicit casts, let's be explicit
```suggestion
        if (!isset($data['content']) || 0 === count($data['content'])) {
```

src/Bridge/Ollama/LlamaModelHandler.php
```diff
@@ -41,7 +42,15 @@ public function request(Model $model, object|array|string $input, array $options
 
     public function convert(ResponseInterface $response, array $options = []): LlmResponse
     {
-        $data = $response->toArray();
+        $data = $response->toArray(false);
```
chr-hertel, line 45: same here, this code silences http error that are not handled below.

```suggestion
        $data = $response->toArray();
```

src/Bridge/OpenAI/Embeddings/ResponseConverter.php
```diff
@@ -20,7 +21,11 @@ public function supports(Model $model, array|string|object $input): bool
 
     public function convert(ResponseInterface $response, array $options = []): VectorResponse
     {
-        $data = $response->toArray();
+        $data = $response->toArray(false);
```
chr-hertel, line 24:
```suggestion
        $data = $response->toArray();
```

src/Bridge/OpenAI/GPT/ResponseConverter.php
```diff
@@ -33,7 +33,7 @@ public function convert(HttpResponse $response, array $options = []): LlmRespons
             return $this->convertStream($response);
         }
 
-        $data = $response->toArray();
+        $data = $response->toArray(false);
```
chr-hertel, line 36:
```suggestion
        $data = $response->toArray();
```

src/Bridge/Replicate/LlamaResponseConverter.php
```diff
@@ -21,7 +22,11 @@ public function supports(Model $model, object|array|string $input): bool
 
     public function convert(HttpResponse $response, array $options = []): LlmResponse
     {
-        $data = $response->toArray();
+        $data = $response->toArray(false);
```
chr-hertel, line 25:
```suggestion
        $data = $response->toArray();
```

src/Bridge/Voyage/ModelHandler.php
```diff
@@ -39,7 +40,11 @@ public function request(Model $model, object|string|array $input, array $options
 
     public function convert(ResponseInterface $response, array $options = []): LlmResponse
     {
-        $response = $response->toArray();
+        $response = $response->toArray(false);
```
chr-hertel, line 43:
```suggestion
        $response = $response->toArray();
```

src/Bridge/Voyage/ModelHandler.php
```diff
@@ -4,6 +4,7 @@
 
 namespace PhpLlm\LlmChain\Bridge\Voyage;
 
+use http\Exception\RuntimeException;
```
chr-hertel, line 7: let's use `PhpLlm\LlmChain\Exception\RuntimeException` instead

src/Bridge/Replicate/LlamaResponseConverter.php
```diff
@@ -4,6 +4,7 @@
 
 namespace PhpLlm\LlmChain\Bridge\Replicate;
 
+use http\Exception\RuntimeException;
```
chr-hertel, line 7: also `PhpLlm\LlmChain\Exception\RuntimeException`

src/Bridge/Ollama/LlamaModelHandler.php
```diff
@@ -4,6 +4,7 @@
 
 namespace PhpLlm\LlmChain\Bridge\Ollama;
 
+use http\Exception\RuntimeException;
```
chr-hertel, line 7: `PhpLlm\LlmChain\Exception\RuntimeException` :)

Iyadhfaleh:
Thank you for your review and your feedbacks @chr-hertel. I applied your suggestions

chr-hertel (approved):
Thank you @Iyadhfaleh

PR #160, opened 2024-12-17
Register tools without `#[AsTool]` attribute

Iyadhfaleh:
The idea is to add tools that haven't `#[AsTool]` attribute like this : 

```php
new ToolBox(new ToolAnalyzer(), ['clock' => $clock]);
```

Tools MUST implement `__invoke` method.


src/Chain/ToolBox/Tool/Names.php
```diff
@@ -0,0 +1,14 @@
...
+enum Names: string
+{
+    case weather = 'weather';
+    case clock = 'clock';
+    case serpapi = 'serpapi';
+    case similarity_search = 'similarity_search';
+    case wikipedia_search = 'wikipedia_search';
+    case wikipedia_article = 'wikipedia_article';
+    case youtube_transcript = 'youtube_transcript';
+}
```
OskarStark, line 14: I think there is no need for an enum

Iyadhfaleh: Thank you for your feedback @OskarStark , I removed it

src/Chain/ToolBox/Tool/ClockWithoutAttribute.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpLlm\LlmChain\Chain\ToolBox\Tool;
+
+use Symfony\Component\Clock\ClockInterface;
+
+final readonly class ClockWithoutAttribute
```
OskarStark, line 9: Such a tool should be added to the tests/ folder, as we don't need to ship two clock tools, doing the exact same thing

Iyadhfaleh: I removed this added example . I already added a fixture test without the attribute . 
The idea was to provide an example of tool without "AsTool" .
I deleted this attribute from clock tool to show that case .

OskarStark: I think we don't need an example for this, but rather a test, so please keep the clock with the attribute and the example and extend `ToolboxTestCase` with a fixture file, shown, that this tool will be registered. thanks

chr-hertel:
Thank you @Iyadhfaleh for your idea to solve this - i'm currently not fully convinced since this is really limited - no description and only `__invoce`. with third party code this is not simply possible.
don't have a better idea myself atm as well.

chr-hertel:
Replaced by #256, see #262 for docs.

Issue #132, opened 2024-10-05
Empower tools to return whatever they want instead of `string`


chr-hertel:
Currently tools can only return `string` which makes sense in the first place since we need to provide the LLM a stringy result, but in the end it would be beneficial for users to just mark basically any public method, that returns something, as tool.

One example would be a `ProductRepository->findByName(string $name): Product`

By default it would be one solution to just call the serializer for `array` and `object` to serialize to `json`.

potentially later we could extend the `AsTool` config to cover the serializer context as well - enabling groups or similar.

PR #161, opened 2024-12-18
Handle `int`, `float`, `array`, `\Stringable` and `\JsonSerializable` return values from tools

OskarStark:
Refs
* 132

### DIFF
--- a/src/Chain/ToolBox/ToolBox.php
+++ b/src/Chain/ToolBox/ToolBox.php
@@ -50,7 +50,13 @@ final class ToolBox implements ToolBoxInterface
         foreach ($this->tools as $tool) {
             foreach ($this->toolAnalyzer->getMetadata($tool::class) as $metadata) {
                 if ($metadata->name === $toolCall->name) {
-                    return $tool->{$metadata->method}(...$toolCall->arguments);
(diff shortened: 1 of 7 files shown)


src/Chain/ToolBox/ToolBox.php
```diff
@@ -50,7 +50,17 @@ public function execute(ToolCall $toolCall): string
...
             foreach ($this->toolAnalyzer->getMetadata($tool::class) as $metadata) {
                 if ($metadata->name === $toolCall->name) {
-                    return $tool->{$metadata->method}(...$toolCall->arguments);
+                    $result = $tool->{$metadata->method}(...$toolCall->arguments);
+
+                    if ($result instanceof \JsonSerializable || is_array($result)) {
+                        return json_encode($result, flags: JSON_THROW_ON_ERROR);
+                    }
+
+                    if (is_integer($result) || is_float($result)) {
```
chr-hertel, line 59: We could add `Stringable` here easily :)

chr-hertel (approved):
Didn't Test, but looks good and pays into #132 :)
Thanks! :)

src/Chain/ToolBox/ToolBox.php
OskarStark: Done

PR #166, opened 2024-12-20
Use `MessageBagInterface` instead of concrete `MessageBag` implementation

OskarStark:


src/Model/Message/MessageBagInterface.php
```diff
@@ -0,0 +1,29 @@
...
+
+declare(strict_types=1);
+
+namespace PhpLlm\LlmChain\Model\Message;
+
+/**
+ * @template-extends \ArrayAccess<int, MessageInterface>
+ * @template-extends \IteratorAggregate<int, MessageInterface>
+ */
+interface MessageBagInterface extends \IteratorAggregate, \ArrayAccess, \Serializable, \Countable, \JsonSerializable
```
OskarStark, line 11:
```php
extends \IteratorAggregate, \ArrayAccess, \Serializable, \Countable,
```
are all from `\ArrayObject`, not sure we need all of them

src/Model/Message/MessageBagInterface.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpLlm\LlmChain\Model\Message;
+
+/**
+ * @template-extends \ArrayAccess<int, MessageInterface>
+ * @template-extends \IteratorAggregate<int, MessageInterface>
```
OskarStark, line 9: not sure about those, but now PHPStan is not complaining anymore 🤷‍♂️

Makefile
```diff
@@ -4,14 +4,14 @@ qa:
...
-	vendor/bin/phpstan
+	vendor/bin/phpstan --memory-limit=-1
 	vendor/bin/phpunit
 
 
 qa-lowest:
 	composer update --prefer-lowest
 	PHP_CS_FIXER_IGNORE_ENV=true vendor/bin/php-cs-fixer fix --diff --verbose
-	vendor/bin/phpstan
+	vendor/bin/phpstan --memory-limit=-1
```
OskarStark, line 14: Not sure why, but without this, I get:
```bash
     Child process error: PHPStan process crashed because it reached configured PHP memory limit: 128M
     Increase your memory limit in php.ini or run PHPStan with --memory-limit CLI option.
      while running parallel worker
     Child process error (exit code 255): PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 262144 bytes) in phar:///Users/oskar.stark/dev/php-llm/llm-chain/vendor/phpstan/phpstan/phpstan.phar/src/Analyser/MutatingScope.php on line 388
     Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 262144 bytes) in phar:///Users/oskar.stark/dev/php-llm/llm-chain/vendor/phpstan/phpstan/phpstan.phar/src/Analyser/MutatingScope.php on line 388
     PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 262144 bytes) in Unknown on line 0
     Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 262144 bytes) in Unknown on line 0
      while running parallel worker
```

src/Model/Message/MessageBagInterface.php
chr-hertel: alternative would be to decouple from that stuff. the only thing we get from that currently is the `$messages[]`

not using `count()` nor iterating over it. array keys are not used as well. at least in my use cases the only thing would be to add an `$messages->add()` method

chr-hertel: WDYT?

chr-hertel: @DZunke any concerns from your end with the breaking change to replace `$messages[] = $message` by `$messages->add($message)`

OskarStark: Separate PR created:
* https://github.com/php-llm/llm-chain/pull/174

chr-hertel:
needs a rebase now

chr-hertel (approved):
Nice one :)

OskarStark:
Rebased and 💚

PR #169, opened 2024-12-21
feat: extend weather tool by forecast and reduced structue

chr-hertel:


tests/Chain/ToolBox/OpenMeteoTest.php
```diff
@@ -0,0 +1,76 @@
...
+            'wind_speed' => '10.7km/h',
+        ];
+
+        static::assertSame($expected, $actual);
+    }
+
+    #[Test]
+    public function forecast(): void
+    {
+        $response = $this->jsonMockResponseFromFile(__DIR__.'/fixtures/openmeteo-forecast.json');
```
OskarStark, line 39: JsonMockResponse::fromFile() is not available?

chr-hertel: got only added with v7.1 - so would break with lowest symfony version

PR #170, opened 2024-12-21
docs: extend tools docs

chr-hertel:


README.md
```diff
@@ -145,6 +145,66 @@ final class CompanyName
...
+}
+```
+
+#### Tool Parameters
+
+LLM Chain generates a JSON Schema representation for all tools in the `ToolBox` based on the `#[AsTool]` attribute and
+method arguments and param comments in the doc block. Additionally, JSON Schema support validation rules, which are
+partially support by LLMs like GPT.
+
+To leverage this, configure the `#[ToolParameter]` attribute on the method arguments of your tool:
```
OskarStark, line 183: Maybe add a link to the class ToolParameter? 

Offtopic, but maybe it's more fail safe to have sth like "#[Minimum]" etc as we can then validate it is only used for integers for example 🤔 to avoid god classes

OskarStark (approved):

README.md
chr-hertel: added a comment below the example code to link to the attribute class, good idea :+1: 

i do get the intention in resolving the mega attribute class, but on the other hand it would make it harder to parse that thing in `ParameterAnalyzer`

PR #173, opened 2024-12-28
feat: override the chain's llm via option

### LINKED ISSUE #172: Override Chain's default model
Make it possible to override the preconfigured llm model while calling the chain via options:

```php
$this->chain->call($messages, ['llm' => $model]);
```

chr-hertel:
closes #172

### DIFF
--- a/src/Chain.php
+++ b/src/Chain.php
@@ -47,20 +47,31 @@ final readonly class Chain implements ChainInterface
      */
     public function call(MessageBag $messages, array $options = []): ResponseInterface
     {
-        $input = new Input($this->llm, $messages, $options);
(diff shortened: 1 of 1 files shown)


src/Chain.php
```diff
@@ -47,20 +47,25 @@ public function __construct(
      */
     public function call(MessageBag $messages, array $options = []): ResponseInterface
     {
-        $input = new Input($this->llm, $messages, $options);
+        if (array_key_exists('llm', $options) && !$options['llm'] instanceof LanguageModel) {
+            throw new InvalidArgumentException('Option "llm" must be an instance of LanguageModel.');
```
OskarStark, line 51: LanguageModel::class to have the FQCN?

OskarStark (approved):

PR #174, opened 2024-12-28
[BC BREAK] `MessageBag` does not depend on `\ArrayObject` anymore

OskarStark:
Follows
* https://github.com/php-llm/llm-chain/pull/166/files#r1898145794

### DIFF
--- a/src/Bridge/Meta/LlamaPromptConverter.php
+++ b/src/Bridge/Meta/LlamaPromptConverter.php
@@ -19,7 +19,7 @@ final class LlamaPromptConverter
         $messages = [];
 
         /** @var UserMessage|SystemMessage|AssistantMessage $message */
-        foreach ($messageBag->getIterator() as $message) {
+        foreach ($messageBag->getMessages() as $message) {
             $messages[] = self::convertMessage($message);
         }
 
--- a/src/Model/Message/MessageBag.php
+++ b/src/Model/Message/MessageBag.php
@@ -4,19 +4,34 @@ declare(strict_types=1);
 
 namespace PhpLlm\LlmChain\Model\Message;
 
-/**
- * @template-extends \ArrayObject<int, MessageInterface>
- */
-final class MessageBag extends \ArrayObject implements \JsonSerializable
+final class MessageBag implements \Countable, \JsonSerializable
 {
+    /**
+     * @var MessageInterface[]
+     */
+    private array $messages;
+
     public function __construct(MessageInterface ...$messages)
     {
-        parent::__construct(array_values($messages));
+        $this->messages = array_values($messages);
+    }
+
+    public function add(MessageInterface $message): void
+    {
+        $this->messages[] = $message;
+    }
+
+    /**
+     * @return MessageInterface[]
+     */
+    public function getMessages(): array
+    {
+        return $this->messages;
     }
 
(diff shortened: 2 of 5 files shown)


src/Bridge/Meta/LlamaPromptConverter.php
```diff
@@ -19,7 +19,7 @@ public function convertToPrompt(MessageBag $messageBag): string
         $messages = [];
 
         /** @var UserMessage|SystemMessage|AssistantMessage $message */
-        foreach ($messageBag->getIterator() as $message) {
+        foreach ($messageBag->messages as $message) {
```
OskarStark, line 22: I like the `getIterator()` more

tests/Model/Message/MessageBagTest.php
```diff
@@ -70,7 +70,7 @@ public function with(): void
         self::assertCount(3, $messageBag);
         self::assertCount(4, $newMessageBag);
 
-        $newMessageFromBag = $newMessageBag[3];
+        $newMessageFromBag = $newMessageBag->messages[3];
```
OskarStark, line 73: I like the previous version more TBH

src/Model/Message/MessageBag.php
```diff
@@ -4,19 +4,26 @@
...
-/**
- * @template-extends \ArrayObject<int, MessageInterface>
- */
-final class MessageBag extends \ArrayObject implements \JsonSerializable
+final class MessageBag implements \Countable, \JsonSerializable
 {
+    /**
+     * @var MessageInterface[]
+     */
+    public array $messages = [];
```
chr-hertel, line 12: if we directly expose the array we cannot ensure any type safety here. everyone could (def including me) skip the `add(...)` method and just write on the array. no matter if `MessageInterface` or not.

too scared, would prefer to be more defensive here and have a `getMessages()` instead

src/Model/Message/MessageBag.php
```diff
@@ -4,19 +4,26 @@
...
+
     public function __construct(MessageInterface ...$messages)
     {
-        parent::__construct(array_values($messages));
+        $this->messages = array_values($messages);
+    }
+
+    public function add(MessageInterface $message): void
```
chr-hertel, line 19: what about
```suggestion
    public function add(MessageInterface ...$messages): void
```
to support adding multiple messages?

src/Model/Message/MessageBag.php
```diff
@@ -73,11 +78,16 @@ public function containsImage(): bool
         return false;
     }
 
+    public function count(): int
+    {
+        return count($this->messages);
+    }
```
chr-hertel, line 84: i like :+1:

chr-hertel:
@OskarStark looks like you have mixed feelings due to missing iterator and array access?

i'm quite positive about this tbh

src/Model/Message/MessageBag.php
OskarStark: done

OskarStark:
Now with the `getMessages()` its more clear as an API, so I am fine 👍

src/Model/Message/MessageBag.php
OskarStark: Hmm the use case is quite unusual I would say.

chr-hertel (approved):
LGTM

PR #175, opened 2024-12-28
feat: add message id, timestamp and other metadata

chr-hertel:


chr-hertel:
closed in favor of #166 and #174

PR #178, opened 2024-12-31
feat: Implement DALL-E image generation for OpenAI Bridge

DZunke:
WDYT? Dall-E 3 and Dall-E 2 have the same endpoint for image generation but differ a bit in input and output. I tried to combine it, even i think on the long run, when OpenAI is not deciding to bring Dall-E 3 nearer to Dall-E 2 features, it would profit from splitting the image generation models in their specific classes for type safety but for now it should be fine, from my PoV. 

It is not urgent from my side to have this feature, as i already utilize it in a slimmer version in my own code. So, if there is something to think about, just let discuss if there are general ideas for improvements here 😸


chr-hertel:
Thanks @DZunke! Looks pretty good in first glance, will give it a try later - thanks :)

OskarStark:
I would also say it looks good, but didn't test it

src/Bridge/OpenAI/DallE.php
```diff
@@ -0,0 +1,34 @@
...
+use PhpLlm\LlmChain\Model\Model;
+
+final readonly class DallE implements Model
+{
+    public const DALL_E_2 = 'dall-e-2';
+    public const DALL_E_3 = 'dall-e-3';
+
+    /** @param array<string, mixed> $options The default options for the model usage */
+    public function __construct(
+        private string $version = self::DALL_E_2,
```
OskarStark, line 16: Should we make version 3 the default?

examples/image-generator-dall-e-2.php
```diff
@@ -0,0 +1,26 @@
...
+    exit(1);
+}
+
+$platform = PlatformFactory::create($_ENV['OPENAI_API_KEY']);
+$llm = new DallE(DallE::DALL_E_2);
+
+$response = $platform->request(
+    model: $llm,
+    input: 'A cartoon-style elephant with a long trunk and large ears.',
+    options: ['response_format' => 'url', 'n' => 2],
```
OskarStark, line 21:
```suggestion
    options: ['response_format' => 'url', 'n' => 2], // generates 2 images
```

examples/image-generator-dall-e-3.php
```diff
@@ -0,0 +1,36 @@
...
+    exit(1);
+}
+
+$platform = PlatformFactory::create($_ENV['OPENAI_API_KEY']);
+$llm = new DallE(DallE::DALL_E_3);
+
+$response = $platform->request(
+    model: $llm,
+    input: 'A cartoon-style elephant with a long trunk and large ears.',
+    options: ['response_format' => 'url', 'n' => 1],
```
OskarStark, line 23:
```suggestion
    options: ['response_format' => 'url', 'n' => 1], // generates 1 image
```

examples/image-generator-dall-e-2.php
```diff
@@ -0,0 +1,26 @@
...
+);
+
```
OskarStark, line 23: Is it possible to output the revised prompt here too?

src/Bridge/OpenAI/DallE/ModelClient.php
```diff
@@ -0,0 +1,66 @@
...
+
+        $images = [];
+        foreach ($response['data'] as $image) {
+            if ('url' === $options['response_format']) {
+                $images[] = new UrlImage($image['url']);
+
+                continue;
+            }
+
+            $images[] = new Base64Image($image['b64_json']);
```
OskarStark, line 61: Maybe add one more example where we use base64 encoded result?

tests/Bridge/OpenAI/DallE/ModelClientTest.php
```diff
@@ -0,0 +1,97 @@
...
+            ->method('request')
+            ->with('POST', 'https://api.openai.com/v1/images/generations', [
+                'auth_bearer' => 'sk-api-key',
+                'json' => [
+                    'model' => DallE::DALL_E_2,
+                    'prompt' => 'foo',
+                    'n' => 1,
+                    'response_format' => 'url',
+                ],
+            ]);
```
OskarStark, line 52: Lets use `MockHttpClient` here directly

tests/Bridge/OpenAI/DallE/ModelClientTest.php
```diff
@@ -0,0 +1,97 @@
...
+
+    #[Test]
+    public function itIsConvertingTheResponse(): void
+    {
+        $httpResponse = self::createStub(HttpResponse::class);
+        $httpResponse->method('toArray')->willReturn([
+            'data' => [
+                ['url' => 'https://example.com/image.jpg'],
+            ],
+        ]);
```
OskarStark, line 66: same here,

tests/Bridge/OpenAI/DallE/ModelClientTest.php
```diff
@@ -0,0 +1,97 @@
...
+    public function itIsConvertingTheResponseWithRevisedPrompt(): void
+    {
+        $emptyPixel = 'iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mNk+M9QDwADhgGAWjR9awAAAABJRU5ErkJggg==';
+
+        $httpResponse = self::createStub(HttpResponse::class);
+        $httpResponse->method('toArray')->willReturn([
+            'data' => [
+                ['b64_json' => $emptyPixel, 'revised_prompt' => 'revised prompt'],
+            ],
+        ]);
```
OskarStark, line 86: same here

tests/Bridge/OpenAI/DallETest.php
```diff
@@ -0,0 +1,34 @@
...
+
+use PhpLlm\LlmChain\Bridge\OpenAI\DallE;
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\Attributes\Small;
+use PHPUnit\Framework\Attributes\Test;
+use PHPUnit\Framework\TestCase;
+
+#[CoversClass(DallE::class)]
+#[Small]
+class DallETest extends TestCase
```
OskarStark, line 15:
```suggestion
final class DallETest extends TestCase
```

tests/Bridge/OpenAI/DallE/UrlImageTest.php
```diff
@@ -0,0 +1,33 @@
...
+
+use PhpLlm\LlmChain\Bridge\OpenAI\DallE\UrlImage;
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\Attributes\Small;
+use PHPUnit\Framework\Attributes\Test;
+use PHPUnit\Framework\TestCase;
+
+#[CoversClass(UrlImage::class)]
+#[Small]
+class UrlImageTest extends TestCase
```
OskarStark, line 15:
```suggestion
final class UrlImageTest extends TestCase
```

tests/Bridge/OpenAI/DallE/ModelClientTest.php
```diff
@@ -0,0 +1,97 @@
...
+
+#[CoversClass(ModelClient::class)]
+#[UsesClass(DallE::class)]
+#[UsesClass(HttpResponse::class)]
+#[UsesClass(HttpClientInterface::class)]
+#[UsesClass(UrlImage::class)]
+#[UsesClass(Base64Image::class)]
+#[UsesClass(GeneratedImagesResponse::class)]
+#[Small]
+class ModelClientTest extends TestCase
```
OskarStark, line 28:
```suggestion
final class ModelClientTest extends TestCase
```

tests/Bridge/OpenAI/DallE/GeneratedImagesResponseTest.php
```diff
@@ -0,0 +1,56 @@
...
+use PHPUnit\Framework\Attributes\Small;
+use PHPUnit\Framework\Attributes\Test;
+use PHPUnit\Framework\Attributes\UsesClass;
+use PHPUnit\Framework\TestCase;
+
+#[CoversClass(GeneratedImagesResponse::class)]
+#[UsesClass(Base64Image::class)]
+#[UsesClass(UrlImage::class)]
+#[Small]
+class GeneratedImagesResponseTest extends TestCase
```
OskarStark, line 20:
```suggestion
final class GeneratedImagesResponseTest extends TestCase
```

tests/Bridge/OpenAI/DallE/Base64ImageTest.php
```diff
@@ -0,0 +1,34 @@
...
+
+use PhpLlm\LlmChain\Bridge\OpenAI\DallE\Base64Image;
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\Attributes\Small;
+use PHPUnit\Framework\Attributes\Test;
+use PHPUnit\Framework\TestCase;
+
+#[CoversClass(Base64Image::class)]
+#[Small]
+class Base64ImageTest extends TestCase
```
OskarStark, line 15:
```suggestion
final class Base64ImageTest extends TestCase
```

chr-hertel:
I guess on detailed level Oskar already made some good remarks, taking one step back my first concerns would be more about OOP slicing and naming.

In general GenAI is about multi modal content input and output and I wonder what your perspective is about promoting the `PhpLlm\LlmChain\Model\Message\Content` one level to `PhpLlm\LlmChain\Model\Content` and also using it with the response instances - we already have an image class there.

src/Bridge/OpenAI/DallE/GeneratedImagesResponse.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpLlm\LlmChain\Bridge\OpenAI\DallE;
+
+use PhpLlm\LlmChain\Model\Response\ResponseInterface;
+
+class GeneratedImagesResponse implements ResponseInterface
```
chr-hertel, line 9: let's rename it to `ImageResponse`

`Generated` is kinda redundant since this is the purpose of the entire thing - also it would be inconsistent with the `TextResponse` since that is not `GeneratedTextsResponse`

and let's promote it from OpenAI Bridge to Model namespace :+1:

src/Bridge/OpenAI/DallE/ModelClient.php
```diff
@@ -0,0 +1,66 @@
...
+use PhpLlm\LlmChain\Platform\ModelClient as PlatformResponseFactory;
+use PhpLlm\LlmChain\Platform\ResponseConverter as PlatformResponseConverter;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+use Symfony\Contracts\HttpClient\ResponseInterface as HttpResponse;
+use Webmozart\Assert\Assert;
+
+/**
+ * @see https://platform.openai.com/docs/api-reference/images/create
+ */
+final readonly class ModelClient implements PlatformResponseFactory, PlatformResponseConverter
```
chr-hertel, line 19: from my point of view totally fine to keep that combined for now :+1:

src/Bridge/OpenAI/DallE/GeneratedImagesResponse.php
OskarStark: ImageSResponse? It can contain many

src/Bridge/OpenAI/DallE.php
DZunke: From my PoV both versions are very comparable, i have just decided to use the lesser costful thingy as default. What would, from your opinion, be a profit to make the newer model the default? Within GPT there is also not the o1 model, which is the newest, the default 🤔

examples/image-generator-dall-e-2.php
DZunke: No. Revised Prompt is a Dall-E 3 only feature. Dall-E 2 is supporting the "n" parameter but Dall-E 3 is rewriting the prompt instead. That is what i mentioned with my comment that the features between the models differ.

src/Bridge/OpenAI/DallE/ModelClient.php
DZunke: I would currently see no profit. The only difference for the user is that the encoded image have to be stored instantly but the links are removed after a period of time by OpenAI. What would be the profit from your PoV?

tests/Bridge/OpenAI/DallETest.php
DZunke: I will change the finals because it seem to be a rule but just out of curiosity ... what is the profit of having the tests being final? If it is a style thingy it should be enforced by the style tooling? Otherwise it is looking like a "stupid work" thingy during a review that one want to not do for themself regulary. What do you think?

src/Bridge/OpenAI/DallE/GeneratedImagesResponse.php
DZunke: Yeah. I would go for `ImagesResponse` as Oskar mentioned 😉 When utilizing Dall-E 2 with the `n` parameter larger then `1` it will be a collection of images.

DZunke:
> I guess on detailed level Oskar already made some good remarks, taking one step back my first concerns would be more about OOP slicing and naming.
> 
> In general GenAI is about multi modal content input and output and I wonder what your perspective is about promoting the `PhpLlm\LlmChain\Model\Message\Content` one level to `PhpLlm\LlmChain\Model\Content` and also using it with the response instances - we already have an image class there.

From my PoV the `ImagesResponse` is very special to Dall-E model because it contains the `revised_prompt` output made by Dall-E 3 as the model rewrites the given user prompt and it is a worth information what prompt was really taken into account for creating the image. WDYT?

examples/image-generator-dall-e-2.php
OskarStark: For me it was hard to guess what `'n'` is doing

src/Bridge/OpenAI/DallE.php
OskarStark: Fine for me if the costs are that much different

tests/Bridge/OpenAI/DallE/ModelClientTest.php
DZunke: Since the last time you noted this it has not changed to error with the message `Symfony\Component\HttpClient\Exception\InvalidArgumentException: MockResponse instances must be issued by MockHttpClient before processing.` when utilizing the `MockResponse`. 

Sorry ... i will not change it to be a more complex testcase just to get the MockClient utilized.

tests/Bridge/OpenAI/DallETest.php
OskarStark: One thing I like is, that it will not popup when you extend a class as it is not possible to extend them. besides that, if I read `final` I know, I don't need to think about another consumer or not. If no `final` there could be some class which extends this class

tests/Bridge/OpenAI/DallE/ModelClientTest.php
DZunke: See https://github.com/php-llm/llm-chain/pull/150#discussion_r1872665463

tests/Bridge/OpenAI/DallE/ModelClientTest.php
DZunke: I still see no advantage in this change. Maybe you are able to describe it to me? Still the only advantage that comes in my mind from the documentation is this:

> However, using [MockResponse](https://github.com/symfony/symfony/blob/7.2/src/Symfony/Component/HttpClient/Response/MockResponse.php) allows simulating chunked responses and timeouts:
See: https://symfony.com/doc/current/http_client.html#http-client-and-responses 

And that is not what i am doing here. I will change it this time for this single test case but in general it would be nice to understand why you are having the opinion it would be better utilized then what i wrote with the mock. The assertions just look a bit different.

tests/Bridge/OpenAI/DallE/ModelClientTest.php
DZunke: Same as https://github.com/php-llm/llm-chain/pull/178/files#r1901833025

examples/image-generator-dall-e-2.php
DZunke: And for me it was what `temperature` is doing 🤣 So you may not have answered my question. What should we generally do?

tests/Bridge/OpenAI/DallETest.php
DZunke: Thanks! So it is "just" a style thing. Like me adding the `final` keyword on classes implementing an interface (that could be mocked / stubbed in tests instead). 

Then we should think about having the cs-fixer or rector doing it to get rid of such monkey comments. Do you know if there would be something available? My experience with php-cs-fixer is not that much.

src/Bridge/OpenAI/DallE/GeneratedImagesResponse.php
chr-hertel: which is basically the same with the GPT completion/chat endpoint.

but it's not consistent at this point - true.

* `VectorResponse` => can handle multiple vectors
* `ToolCallResponse` => can handle multiple tool calls
* `ChoiceResponse` => can handle multiple choices <= this is where it feels inconsistent because that should be feature already of `TextResponse` and `ToolCallResponse`

but anyhow, `ImagesResponse` would be the first with a plural

---

long story short - i'd say if plural than for all :D

chr-hertel:
> is very special to Dall-E model because it contains the revised_prompt output made by Dall-E 3 as the model rewrites the given user prompt

fair, and we can generalize later - but on the first research this is pretty unique to DallE, true

examples/image-generator-dall-e-2.php
chr-hertel: in general `$options` are a mixture of keys that a) refer to features introduced by processors, e.g. `llm` or `output_structure`, or options provided by the underlying model, e.g. `n`.

i guess that should be documented better in the readme, see https://github.com/php-llm/llm-chain/pull/183

anyhow, for examples i think it is just helpful - if options are used - to add a comment :+1: no harm

examples/image-generator-dall-e-3.php
chr-hertel: in this case i would rather drop the options since it's repeating the defaults, right?

tests/Bridge/OpenAI/DallE/ModelClientTest.php
chr-hertel: same, too many bad experience and refactoring pain with mocks - even if i feel like the MockHttpClient is that smooth.

thanks for adopting! :+1:

src/Bridge/OpenAI/DallE/GeneratedImagesResponse.php
OskarStark: Let's go with singular then

src/Bridge/OpenAI/DallE/ModelClient.php
OskarStark: ok

OskarStark (approved):

examples/image-generator-dall-e-2.php
DZunke: I have decided to remove the default options from the model, instead made some adjustments to the examples with more comments to what is meant with this option.

examples/image-generator-dall-e-3.php
DZunke: See https://github.com/php-llm/llm-chain/pull/178#discussion_r1904282413

src/Bridge/OpenAI/DallE/GeneratedImagesResponse.php
DZunke: Renamed to singular  👍🏼

chr-hertel (approved):
Thanks @DZunke - nice addition :)

PR #181, opened 2025-01-02
Adjust test

OskarStark:


tests/Model/Message/MessageBagTest.php
```diff
@@ -103,18 +103,23 @@ public function withoutSystemMessage(): void
         $messageBag = new MessageBag(
             Message::forSystem('My amazing system prompt.'),
             Message::ofAssistant('It is time to sleep.'),
+            Message::forSystem('A system prompt in the middle.'),
             Message::ofUser('Hello, world!'),
+            Message::forSystem('Another system prompt at the end'),
```
chr-hertel, line 108: i feel like we're running a bit in circles with this :laughing: in my understanding it is wrong to use multiple system/developer messages. can you point me again to your use-case so i can better understand what you're trying to achieve?

tests/Model/Message/MessageBagTest.php
```diff
@@ -103,18 +103,23 @@ public function withoutSystemMessage(): void
...
-        $messageFromNewBag = $newMessageBag->getMessages()[0];
+        $assistantMessage = $newMessageBag->getMessages()[0];
+        self::assertInstanceOf(AssistantMessage::class, $assistantMessage);
+        self::assertSame('It is time to sleep.', $assistantMessage->content);
 
-        self::assertInstanceOf(AssistantMessage::class, $messageFromNewBag);
-        self::assertSame('It is time to sleep.', $messageFromNewBag->content);
+        $userMessage = $newMessageBag->getMessages()[1];
+        self::assertInstanceOf(UserMessage::class, $userMessage);
+        self::assertSame('Hello, world!', $userMessage->content[0]->text);
```
OskarStark, line 122: Any clue how to get rid of this PHPStan error @chr-hertel ?

DZunke: I think it would be no fautl to add phpstan-phpunit extension. Then it should detect that you checked the instance of before accessing the content. Or you have to add an instanceof assertion for the index key of the content[0].

tests/Model/Message/MessageBagTest.php
chr-hertel: hehe, just saying^^

tests/Model/Message/MessageBagTest.php
chr-hertel: in this case the assert was enough :+1:

chr-hertel (approved):

PR #184, opened 2025-01-03
feat: empower input processor to change LLM and MessageBag

chr-hertel:


OskarStark (approved):

OskarStark:
It if I get this PR right, this is just the foundation to be able to register such InputProcessor, right?

chr-hertel:
> But if I get this PR right, this is just the foundation to be able to register such InputProcessor, right?

Regarding the system prompt thingy? yes

Issue #171, opened 2024-12-21
Configure a default system prompt for a chain


OskarStark:
This could look like this in the bundle:
```yaml
    chain:
        default:
            platform: 'llm_chain.platform.azure.default'
            model:
                name: 'GPT'
                version: 'gpt-4o'
                options:
                    temperature: 0.8
            tools: false
            system_prompt: |
                You are a helpful assistant.
                You are here to provide information and help users find the information they need.
                You are polite and professional, and you always provide accurate information.
```

OskarStark:
@DZunke what do you think?

DZunke:
Hey Oskar, for a default system prompt this would surely fit as a configuration. But i do not utilize the bundle because i have a lot more runtime changes then only the system prompt, in result i have my own chain factory working. 

So it is hard for me to give feedback to the bundle configuration in general, sorry 🙈 

But more generally I am sure for some use cases a general default system prompt is a nice idea but i would fully pull this issue to the bundle and extend the functionality there. I do not see a reason to have the library itself having a mechanism for such defaults as already the readme states out how it works with a new message bag with a default system prompt by the own code.

chr-hertel:
i totally agree that this would be super nice, but it is indeed somehow tricky since the `MessageBag` is provided to the chain and not originating from the chain. however i was also wondering if there is a clever way to abstract the Chat classes of the demo project: https://github.com/php-llm/symfony-demo/blob/main/src/Blog/Chat.php

OskarStark:
What about giving this system prompt via the options to the call() ? As we have message value objects we could check if this is already included in the message bag and otherwise prepend the system prompt.

The code part needs to be done in the lib for this, the bundle config will then follow

chr-hertel:
Ah nice - it could even be an Input processor

chr-hertel:
Def open for a PR :)

chr-hertel:
This PR should be helpful in implementing an input processor to inject a system message
https://github.com/php-llm/llm-chain/pull/184

chr-hertel:
Solved with v0.13 or lib and bundle.

PR #186, opened 2025-01-04
feat: add system prompt processor

chr-hertel:
Foundation for #171 - everything else need to go to the bundle i guess.

### DIFF
--- a/src/Chain/InputProcessor/SystemPromptInputProcessor.php
+++ b/src/Chain/InputProcessor/SystemPromptInputProcessor.php
@@ -1,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpLlm\LlmChain\Chain\InputProcessor;
+
+use PhpLlm\LlmChain\Chain\Input;
+use PhpLlm\LlmChain\Chain\InputProcessor;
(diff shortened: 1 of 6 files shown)


src/Chain/InputProcessor/SystemPromptInputProcessor.php
```diff
@@ -0,0 +1,26 @@
...
+    public function __construct(
+        private string $systemPrompt,
+    ) {
+    }
+
+    public function processInput(Input $input): void
+    {
+        $messages = $input->messages;
+
+        if (null === $messages->getSystemMessage()) {
```
OskarStark, line 22: We should add a log here at least. I also thought about throwing, but this way you are not able to overwrite the system prompt in certain situations

src/Chain/InputProcessor/SystemPromptInputProcessor.php
```diff
@@ -0,0 +1,26 @@
...
+            $input->messages = $messages->prepend(new SystemMessage($this->systemPrompt));
```
OskarStark, line 23: Message::ofSystem?

OskarStark (approved):

PR #187, opened 2025-01-04
feat: add tool call event

### LINKED ISSUE #168: Let the tool decide if the chain should be called again with the return value
In my case I craft a dedicated response like:
```json
{
  "foo": "bar",
}
```
and I don't want to feed it back to the LLM for another roundtrip.

That happens here (in Line 55):
https://github.com/php-llm/llm-chain/blob/8365e13aa892abfc3a2c54d2067c82cf8e30e830/src/Chain/ToolBox/ChainProcessor.php#L46-L55

I think we should not make it configurable like "Tool A will always feed back to the API", because it can depend on runtime. Maybe a `ToolResponse` class as an additional allowed return type?

Dummy-code

```php
new class ToolResponse {
    value: string, float etc.
    callLlm: yes/no
}
```

chr-hertel:
Closes #168


chr-hertel:
Tests currently failing since i moved tool result conversion into the chain processor, but would that be a way to go for you @OskarStark ?

```php
final class SpecificToolCallSubscriber implement EventSubscriberInterface
{
    public static function getSubscribedEvents()
    {
        return [
            ToolCallExecuted::class => 'handleToolCall',
        ];
    }

    public function handleToolCall(ToolCallExecuted $event): void
    {
        foreach ($event->toolCallResults as $toolCallResult) {
            if (str_starts_with($toolCallResult->toolCall->name, 'weather_')) {
                $event->response = new StructuredResponse($toolCallResult->result);
            }
        }
    }
}
```

OskarStark:
Uhhhhhhh that looks nice 😍 will have a detailed look when on my Mac 💻👍🏻

Thanks for taking time

examples/toolbox-weather-event.php
```diff
@@ -23,12 +26,22 @@
...
+$openMeteo = new OpenMeteo(HttpClient::create());
+$toolBox = new ToolBox(new ToolAnalyzer(), [$openMeteo]);
+$eventDispatcher = new EventDispatcher();
+$processor = new ChainProcessor($toolBox, eventDispatcher: $eventDispatcher);
 $chain = new Chain($platform, $llm, [$processor], [$processor]);
 
-$messages = new MessageBag(Message::ofUser('How is the weather currently in Berlin? And how about tomorrow?'));
+// Add tool call result listener to enforce chain exits direct with structured response for weather tools
+$eventDispatcher->addListener(ToolCallExecuted::class, function (ToolCallExecuted $event): void {
+    foreach ($event->toolCallResults as $toolCallResult) {
```
OskarStark, line 37: The event name is singular and then you iterate over all toolCallResults (plural), that feels weird. Shouldn't we throw an event for every ToolCall which was executed?

Otherwise rename it to `ToolCallsExecuted`?

Or am I missing sth.?

src/Chain/ToolBox/ToolResultConverter.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpLlm\LlmChain\Chain\ToolBox;
+
+final readonly class ToolResultConverter
```
OskarStark, line 7: This will only be used if no subscriber modified the `$event->response`, right?

OskarStark:
I like it 👍 Looks like I can achieve my goals with it 🕺

src/Chain/ToolBox/ToolResultConverter.php
chr-hertel: no, it's executed before the event is thrown, see https://github.com/php-llm/llm-chain/pull/187/files#diff-a9c414f9addf94c57860c2cdb4f858ef784e2c7a9e958649df6e66554ef7c434R58

examples/toolbox-weather-event.php
chr-hertel: you're right. should be plural

src/Chain/ToolBox/ToolResultConverter.php
OskarStark: Works for me, but wouldn't it open up the possibilities if we would have access to the raw result from the tool in the subscriber?

OFC the subscriber would then be forced to "return" a `string` to the event, not sure it's great?

but yeah, lets go with it for now and see how things are going

OskarStark (approved):

src/Chain/ToolBox/ToolResultConverter.php
OskarStark: Wait...

OskarStark: I have a tool where I would like to return it to the chain if "No products found" text is returned, because it should maybe translated to the conversation which is in Turkish, but if I return a `ProductsResponse` (\JsonSerializable) I want to intercept the chain.

Right now I end up with a string in my subscriber, where I need to decide intercept or not. Here the raw result (`ProductsResponse`) would be helpful for the determination

chr-hertel: the converted result is only used for the message. you should get the raw result in the event, don't  you?

OskarStark: Indeed, sorry

PR #189, opened 2025-01-05
feat: audio input support

chr-hertel:


src/Model/Message/MessageBagInterface.php
chr-hertel, line 1: this approach doesn't scale well

OskarStark (approved):

PR #192, opened 2025-01-06
Add possibility to react to content filter violation

OskarStark:
I am unsure about the logic, as this structure is only working for OpenAPI:
CleanShot 2025-01-06 at 13 53 44@2x

Any ideas @chr-hertel ?


OskarStark:
Yes why not

OskarStark:
I pushed an update, is this what you had in mind? Code untested right now

src/Bridge/OpenAI/GPT/ResponseConverter.php
```diff
@@ -33,7 +34,17 @@ public function convert(HttpResponse $response, array $options = []): LlmRespons
             return $this->convertStream($response);
         }
 
-        $data = $response->toArray();
+        try {
+            $data = $response->toArray();
+        } catch (\Throwable $e) {
```
chr-hertel, line 39: in this case maybe catching only `ClientExceptionInterface`?

OskarStark: done

PR #194, opened 2025-01-06
JSON Schema needs all properties as "required"

OskarStark:
Based on:
CleanShot 2025-01-06 at 18 43 48@2x

Link: https://platform.openai.com/docs/guides/function-calling#example

I am not sure if this is also the case for other platforms/models

### DIFF
--- a/src/Chain/StructuredOutput/SchemaFactory.php
+++ b/src/Chain/StructuredOutput/SchemaFactory.php
@@ -66,11 +66,7 @@ final class SchemaFactory
 
             // Add property schema to main schema
             $schema['properties'][$propertyName] = $propertySchema;
-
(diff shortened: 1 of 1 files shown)


chr-hertel:
not sure i get this - have you been running into issues? this is for structured output after calling tools?

OskarStark:
> not sure i get this - have you been running into issues? this is for structured output after calling tools?

Ah no not here, but in my other project with my code, but it was more about the schema for a tool description

chr-hertel:
but still, if you want to change the schema provided for tools, you're patching the wrong class here - anyhow, i will add some tests and maybe unify that parsing stuff :thinking:

OskarStark:
lets close it here

PR #195, opened 2025-01-06
Allow `null` as type in schema

OskarStark:
Otherwise I experienced, if it has no info, it starts to imagine a value, just to fulfill the schema


chr-hertel:
Arrays could be nullable as well and I wonder if that maybe better to have it in `SchemaFactory::buildSchema:61` - to have that logic once instead of having the if for every type again

OskarStark:
Will check later and yes I forgot about arrays

OskarStark:
PR updated, like it!

chr-hertel (approved):
Merci :pray:

PR #197, opened 2025-01-22
Use retryable HTTP client

odinuv:


chr-hertel:
i think you can easily achieve that by injecting the `RetryableHttpClient` into the `create` method.

are you using the bundle or the lib standalone?

odinuv:
Thanks for the reply! I'm using the bundle, I didn't seem to work, but I might have been doing something wrong for sure. Right now I need to build a test image of our app to check that it will actually help resolve our issue. If so, I'll surely try to find a better way to do this.

odinuv:
So my colleague tried this again and it works by injecting the client - e.g. setting in Symfony config. 

```
framework:
    http_client:
        default_options:
            timeout: 300
            max_duration: 300
            retry_failed:
                max_retries: 3
                delay: 1000
                multiplier: 2.0
                jitter: 0.1

```

Closing this as not needed.

PR #202, opened 2025-01-24
Support descriptions from @param

OskarStark:


OskarStark:
@chr-hertel this shows the problem, but I have no clue how to extract the description from `@param` 🤔 

cc @DZunke

chr-hertel:
had a look and failed as well, and i guess that's because

a) the param tag belongs to methods

meaning it should look like
```php
final class UserWithAtParamAnnotation
{
    /**
     * @param string $name The name of the user in lowercase
     */
    public function __construct(
        public int $id,
        public string $name,
        public \DateTimeInterface $createdAt,
        public bool $isActive,
        public ?int $age = null,
    ) {
    }
}
```

which leads to b) the property component doesn't care about method params from constructor, is that possible?

meaning those param infos are not extracted by Symfony's `PropertyType` component

OskarStark:
That is possible

chr-hertel:
so, after further investigation it doesn't even work with https://github.com/symfony/symfony/pull/57632

so decouple from property-info and doing it ourselves would be the answer :grimacing:

OskarStark:
Needs
* https://github.com/symfony/symfony/pull/59540

OskarStark:
Superseded by #212 

Thanks @chr-hertel

PR #203, opened 2025-01-31
Added support for OpenRouter & Generic Models

rglozman:
Added support for OpenRouter which hosts over 300 models. This includes support for models from every provider into one easy to use API.

I created an example using Gemini (free model hosted by google), updated the .env, and the README.


chr-hertel:
Hi, thanks for the idea and I like the `GenericModel` approach for this platform :+1: 

Was just creating an API key for openrouter, was running the example you pushed and got this error:
```
php examples/chat-gemini-openrouter.php 
PHP Fatal error:  Uncaught PhpLlm\LlmChain\Exception\RuntimeException: Response does not contain message in /.../llm-chain/src/Bridge/OpenRouter/Client.php:54
Stack trace:
#0 /.../llm-chain/src/Model/Response/AsyncResponse.php(33): PhpLlm\LlmChain\Bridge\OpenRouter\Client->convert()
#1 /.../llm-chain/src/Chain.php(75): PhpLlm\LlmChain\Model\Response\AsyncResponse->unwrap()
#2 /.../llm-chain/examples/chat-gemini-openrouter.php(26): PhpLlm\LlmChain\Chain->call()
#3 {main}
  thrown in /home/christopher/Projects/LlmChain/llm-chain/src/Bridge/OpenRouter/Client.php on line 54
```

does it work for you?

---

btw, this should fix the codestyle issues in the pipeline:
```
PHP_CS_FIXER_IGNORE_ENV=1 vendor/bin/php-cs-fixer fix
```
or to run the entire quality pipeline locally:
```
make qa
```

rglozman:
Could you add var_dump($data) just before line 54 in src/Bridge/OpenRouter/Client.php and run again? Inside the if statement just before the "throw new". Could be a bunch of different errors (auth, out of credits, servers down, etc...). I am working on a better way to report errors to the user (for all providers), I will make it a separate pull request.

Yes I will run the linter, thanks for the heads up.

chr-hertel:
You are right, just tried it again with `dump` on the response. First time it worked, but second time it was rejected due to rate limit.

The error handling is something to improve in general. I would love to also have a more generalistic approach, but also the error responses depend on the specific model and platform. so a combination could be worth exploring :thinking: like a better default error handling and an extension point per platform/model that enables bridges to fine-tune specifics of error handling.

chr-hertel:
So, all in all, thanks for the PR, happy to merge once the pipeline is green :) :+1:

rglozman:
Should be good to merge now 👍

chr-hertel (approved):
Thanks @rglozman! :+1:

PR #206, opened 2025-02-05
chore: remove phpdocumentor conflict

chr-hertel:
Following up on #143 - cannot reproduce the issue anymore :thinking: 

@DZunke do you still recall what happened?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,8 +43,7 @@
         "symfony/var-dumper": "^6.4 || ^7.1"
     },
     "conflict": {
-        "mongodb/mongodb": "<1.20",
-        "phpdocumentor/reflection-docblock": ">=5.6"
+        "mongodb/mongodb": "<1.20"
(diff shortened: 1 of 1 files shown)


DZunke:
@chr-hertel  was nothing big. After the library was updated i had the problem during a Pull Request building when i executed the testsuite. As mentioned in #143 the tets `ChainProcessorTest` failed with the error there. 

In december they published a new patch for compatibility with phpstan 2, so maybe the lexer is now working as the error came from a phpstan library. 

https://github.com/phpDocumentor/ReflectionDocBlock/releases/tag/5.6.1

Maybe the conflict is not needed anymore after mid of december.

chr-hertel:
Thanks, yes, saw that and was just wondering because i couldn't even reproduce with 5.6.0 but nvm - looks like it's gone for good :)

PR #208, opened 2025-02-06
feat: chat + message store

chr-hertel:


src/Chat/MessageStore/CacheStore.php
```diff
@@ -0,0 +1,41 @@
...
+
+use PhpLlm\LlmChain\Chat\MessageStoreInterface;
+use PhpLlm\LlmChain\Model\Message\MessageBag;
+use Psr\Cache\CacheItemPoolInterface;
+
+final readonly class CacheStore implements MessageStoreInterface
+{
+    public function __construct(
+        private CacheItemPoolInterface $cache,
+        private string $cacheKey = 'messages',
```
OskarStark, line 15: using it in parallel would overwrite the messages of another conversation

src/ChatInterface.php
OskarStark, line 1: like 👍

src/Chat/MessageStore/CacheStore.php
chr-hertel: would you be more strict and enforce the key is injected?

examples/persistent-chat.php
```diff
@@ -0,0 +1,34 @@
...
+$llm = new GPT(GPT::GPT_4O_MINI);
+
+$chain = new Chain($platform, $llm);
+$chat = new Chat($chain, new InMemoryStore());
+
+$messages = new MessageBag(
+    Message::forSystem('You are a helpful assistant. You only answer with short sentences.'),
+);
+
+$chat->start($messages);
```
OskarStark, line 30:
```suggestion
$chat->initialize($messages);
```
? Bootstrap? Configure?

OskarStark: initiate is nice

chr-hertel: `initiate` it is :+1:

PR #209, opened 2025-02-07
feat: add support for Anthropic tool use

soukicz:
I have added Anthropic tool use support but I cannot figure out how to preserve text content that is sent with tool use request. I have tried to copy OpenAI function calling but that seems to loose text content during function call too.

### DIFF
--- a/src/Bridge/Anthropic/Claude.php
+++ b/src/Bridge/Anthropic/Claude.php
@@ -54,6 +54,6 @@ final readonly class Claude implements LanguageModel
 
     public function supportsToolCalling(): bool
     {
-        return false; // it does, but implementation here is still open.
+        return true;
     }
 }
--- a/src/Bridge/Anthropic/ModelHandler.php
+++ b/src/Bridge/Anthropic/ModelHandler.php
@@ -4,12 +4,18 @@ declare(strict_types=1);
 
 namespace PhpLlm\LlmChain\Bridge\Anthropic;
 
+use PhpLlm\LlmChain\Chain\ToolBox\Metadata;
 use PhpLlm\LlmChain\Exception\RuntimeException;
+use PhpLlm\LlmChain\Model\Message\AssistantMessage;
 use PhpLlm\LlmChain\Model\Message\MessageBagInterface;
+use PhpLlm\LlmChain\Model\Message\MessageInterface;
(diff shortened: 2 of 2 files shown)


chr-hertel:
Hey, thanks for working on this!

Can you describe what you mean with "loose text content" please? Not sure i can follow.

An example with code and expectated behavior would be great

soukicz:
I was working mostly with Anthropic and it might be specific only to them. There is always included chain of thought in tool use request:

(it is the `<thinking>` part)
```
curl https://api.anthropic.com/v1/messages \
     --header "x-api-key: $ANTHROPIC_API_KEY" \
     --header "anthropic-version: 2023-06-01" \
     --header "content-type: application/json" \
     --data \
'{
    "model": "claude-3-5-sonnet-20241022",
    "max_tokens": 1024,
    "tools": [
        {
            "name": "get_weather",
            "description": "Get the current weather in a given location",
            "input_schema": {
                "type": "object",
                "properties": {
                    "location": {
                        "type": "string",
                        "description": "The city and state, e.g. San Francisco, CA"
                    },
                    "unit": {
                        "type": "string",
                        "enum": ["celsius", "fahrenheit"],
                        "description": "The unit of temperature, either \"celsius\" or \"fahrenheit\""
                    }
                },
                "required": ["location"]
            }
        }
    ],
    "messages": [
        {
            "role": "user",
            "content": "What is the weather like in San Francisco?"
        },
        {
            "role": "assistant",
            "content": [
                {
                    "type": "text",
                    "text": "<thinking>I need to use get_weather, and the user wants SF, which is likely San Francisco, CA.</thinking>"
                },
                {
                    "type": "tool_use",
                    "id": "toolu_01A09q90qw90lq917835lq9",
                    "name": "get_weather",
                    "input": {
                        "location": "San Francisco, CA",
                        "unit": "celsius"
                    }
                }
            ]
        },
        {
            "role": "user",
            "content": [
                {
                    "type": "tool_result",
                    "tool_use_id": "toolu_01A09q90qw90lq917835lq9",
                    "content": "15 degrees"
                }
            ]
        }
    ]
}'
```


https://docs.anthropic.com/en/docs/build-with-claude/tool-use#single-tool-example

chr-hertel (approved):
Let's merge this in finally - and build up on this.

Sorry for the delay and thanks for your work! :pray:

PR #210, opened 2025-02-10
Add `o3-mini` and `o3-mini-high` models

OskarStark:


src/Bridge/OpenAI/GPT.php
```diff
@@ -17,6 +17,8 @@ final class GPT implements LanguageModel
     public const GPT_4O_AUDIO = 'gpt-4o-audio-preview';
     public const O1_MINI = 'o1-mini';
     public const O1_PREVIEW = 'o1-preview';
+    public const O3_MINI = 'o3-mini';
+    public const O3_MINI_HIGH = 'o3-mini-high';
```
chr-hertel, line 21: can't find that here:
https://platform.openai.com/docs/models#o3-mini

chr-hertel (approved):

PR #212, opened 2025-02-15
refactor: centralize json schema generation

chr-hertel:
Centralizing JSON Schema generation for Output Structures and Tools


README.md
```diff
@@ -213,17 +213,17 @@ final class MyTool
      * @param int    $number The number of an object
      */
     public function __invoke(
-        #[ToolParameter(pattern: '/([a-z0-1]){5}/')]
+        #[With(pattern: '/([a-z0-1]){5}/')]
```
OskarStark, line 216: I propose to use an extra PR for the renaming to `#[With]`

OskarStark: Besides that, we can also add an internal map, so that you can only use minimum and maximum with an int for example + we can throw an exception if the properties are not typed, or do we already?

src/Chain/StructuredOutput/ResponseFormatFactory.php
```diff
@@ -4,12 +4,17 @@
 
 namespace PhpLlm\LlmChain\Chain\StructuredOutput;
 
+use PhpLlm\LlmChain\Chain\JsonSchema\Factory;
+
 use function Symfony\Component\String\u;
 
+/**
+ * @internal
```
OskarStark, line 12: why?

README.md
chr-hertel: so just move it first, rename it later you mean?

OskarStark: Updated my comment, in case you missed it

chr-hertel: Yeah, should think of getting this smaller anyways, true. Refactoring wasn't as successful as hoped for

src/Chain/StructuredOutput/ResponseFormatFactory.php
chr-hertel: removed :+1:

tests/Fixture/Tool/ToolWithToolParameterAttribute.php
```diff
@@ -57,7 +57,7 @@ public function __invoke(
             maxProperties: 10,
             dependentRequired: true,
         )]
-        object $shippingAddress,
+        string $shippingAddress,
```
OskarStark, line 60: why?

chr-hertel: couldn't find a way to translate that to a meaningful schema.

we can't grab any information about the subitems here - or do you see a way?

OskarStark: but it worked before, right?

composer.json
```diff
@@ -17,14 +17,15 @@
...
+        "phpstan/phpdoc-parser": "^2.1",
         "psr/cache": "^3.0",
         "psr/log": "^3.0",
         "symfony/clock": "^6.4 || ^7.1",
         "symfony/http-client": "^6.4 || ^7.1",
         "symfony/property-access": "^6.4 || ^7.1",
         "symfony/property-info": "^6.4 || ^7.1",
         "symfony/serializer": "^6.4 || ^7.1",
-        "symfony/type-info": "^6.4 || ^7.1",
+        "symfony/type-info": "^7.2.3",
```
OskarStark, line 28: do we really need 7.2.3 or can we use 7.2 here?

chr-hertel: yup, unfortunately

tests/Fixture/StructuredOutput/UserWithConstructor.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpLlm\LlmChain\Tests\Fixture\StructuredOutput;
+
+final class UserWithConstructor
+{
+    /**
+     * @param string $name The name of the user in lowercase
```
OskarStark, line 10: ❤️

OskarStark (approved):

composer.json
OskarStark: Ok

PR #213, opened 2025-02-15
feat: introduce optional fault tolerant toolbox

chr-hertel:
the idea is to feedback to the llm that the used tool is not available and try again.

had that a few times now.

**[BC BREAK] Change of toolbox exception namespaces**


OskarStark:
I have mixed feelings about the solution, I would rather throw a dedicated event and let the user chose the solution, WDYT?

We can also add something to the docs about it

chr-hertel:
Okay okay okay :D - couldn't wrap my head around an exception tho and went with decoration - also cleaner i guess :)

OskarStark:
With a fast review it looks good, let's merge it

OskarStark (approved):

PR #216, opened 2025-02-18
Fix: support finish reason "length"

odinuv:
The reason is real https://community.openai.com/t/bug-in-api-response-finish-reason-field/287212 and is even mentioned here https://github.com/php-llm/llm-chain/blob/5da000ec0fb4fcc4d78399e40eac05506cfff6fc/src/Bridge/OpenAI/GPT/ResponseConverter.php#L159 Plus added some tests for the ResponseConverter


chr-hertel:
Hi @odinuv & welcome to the project! :)

Good catch and I'm happy to merge as soon as the pipeline is green - I guess `php vendor/bin/php-cs-fixer fix` should help with that

chr-hertel (approved):

chr-hertel:
Thank you @odinuv :pray:

PR #218, opened 2025-02-18
refactor: gpt stream conversion

chr-hertel:
Easier to read? WDYT @DZunke?


DZunke:
Yeah! Moving those methods to a single loop should improve the readability here for the Response Converter - also performed a test execution of a stream locally. Thanks!

PR #225, opened 2025-02-25
chore: bump dev dependencies

chr-hertel:


composer.json
```diff
@@ -27,10 +27,10 @@
     "require-dev": {
         "codewithkyrian/chromadb-php": "^0.2.1",
         "mongodb/mongodb": "^1.20",
-        "php-cs-fixer/shim": "^3.64",
+        "php-cs-fixer/shim": "^3.70",
         "phpstan/phpstan": "^1.12",
         "phpstan/phpstan-webmozart-assert": "^1.2",
-        "phpunit/phpunit": "^11.3",
+        "phpunit/phpunit": "^11.5",
```
OskarStark, line 33: 12 is already a thing, isn't it?

chr-hertel: Would need to drop 8.2 support then

OskarStark: Uh no, all good then!

PR #229, opened 2025-03-03
Add Oskar as author

OskarStark:
Feel free to close, if you don't agree, no problem


chr-hertel (approved):
should have done that a while ago :)

PR #230, opened 2025-03-03
Require PHPStan and Rector 2

OskarStark:


OskarStark:
Looks like we can ignore those errors

chr-hertel (approved):

PR #233, opened 2025-03-06
Ability to add the tool definitions to the system prompt

OskarStark:


src/Chain/InputProcessor/SystemPromptInputProcessor.php
```diff
@@ -28,6 +35,26 @@ public function processInput(Input $input): void
...
 
-        $input->messages = $messages->prepend(Message::forSystem($this->systemPrompt));
+        $message = $this->systemPrompt;
+
+        if ($this->includeToolDefinitions) {
+            $tools = implode(PHP_EOL.PHP_EOL, array_map(
+                fn (Metadata $tool) => <<<TOOL
+## {$tool->name}
+{$tool->description}
+TOOL,
```
chr-hertel, line 45: please fix the indentation, see https://wiki.php.net/rfc/flexible_heredoc_nowdoc_syntaxes

chr-hertel:
other than that it looks good to me, but would also need a check with the bundle i guess

OskarStark:
I think a dedicated processor would be better tomatoes dealing with optional dependencies

src/Chain/InputProcessor/SystemPromptInputProcessor.php
OskarStark: See 
* #236

chr-hertel:
> Not sure if we need this example here 🤔

yeah, wouldn't say it's needed, but documentation

src/Chain/InputProcessor/SystemPromptInputProcessor.php
```diff
@@ -14,8 +16,13 @@
 {
     public function __construct(
         private string $systemPrompt,
+        private ?ToolBoxInterface $toolBox = null,
```
OskarStark, line 19: @chr-hertel is it to implicit to add the definitions, once a toolbox is available?

chr-hertel: why else would you inject a toolbox?

chr-hertel (approved):
:muscle: :sunglasses:

PR #234, opened 2025-03-06
[PHP-CS-Fixer] Enable `heredoc_indentation` with `start_plus_one` rule

OskarStark:
See https://cs.symfony.com/doc/rules/whitespace/heredoc_indentation.html

This is the default value for this fixer


chr-hertel (approved):
yup, we should automate this :pray: :+1:

PR #237, opened 2025-03-06
ci: add PR validation and auto labeling

chr-hertel:


chr-hertel:
@OskarStark maybe in combination with an action and also adding labels by default?

chr-hertel:
@llupa are you using something similar?

llupa:
> @llupa are you using something similar?

This action looks so much better than what we did back in 2023, I might use it myself :D (with fewer types for the API project you saw last week).

PR #239, opened 2025-03-06
refactor: rename tool param attribute

chr-hertel:
Pulled out of #212 for a more neutral name also usable with DTO for structured output


tests/Fixture/Tool/ToolWithToolParameterAttribute.php
```diff
@@ -21,45 +21,45 @@ final class ToolWithToolParameterAttribute
      * @param object $shippingAddress  The shipping address given to the tool
      */
     public function __invoke(
-        #[ToolParameter(
+        #[With(
```
OskarStark, line 24: As this is now quite short, what about putting the entire attribute in one line?

OskarStark (approved):

PR #248, opened 2025-03-12
refactor: ease toolbox instatiation

chr-hertel:
Extracted from #249


src/Chain/ToolBox/ToolBox.php
```diff
@@ -33,6 +33,11 @@ public function __construct(
         $this->tools = $tools instanceof \Traversable ? iterator_to_array($tools) : $tools;
     }
 
+    public static function withTools(object ...$tools): self
```
OskarStark, line 36:
```suggestion
    public static function create(object ...$tools): self
```
?

OskarStark (approved):

PR #255, opened 2025-03-14
refactor: remove tool metadata class awareness

chr-hertel:
Especially with MCP the tool doesn't have to be a class and it was actually not used.


chr-hertel:
closed in favor of #260

PR #256, opened 2025-03-14
feat: introduce chain and memory tool metadata factory

### LINKED ISSUE #134: Override a Tools configuration
Especially when working with built-in or potentially other 3rd Party Tools it might be beneficial to override the pre-defined configuration provided via the `#[AsTool] attribute.
Also considering a 3rd party service class that doesn't have that attribute but users want to use as tool

chr-hertel:
* allows to use tools without attributes, but still leverages reflection for parameters
* allows to override tools configured with attributes with an explicit config

closes #134 & replaces #160

foundation for using third party classes as tools and registering tools dynamically via MCP


chr-hertel:
also the possibility is needed to overwrite params and get rid off the `method` in metadata object.

PR #260, opened 2025-03-19
refactor: introduce execution reference in tool metadata

chr-hertel:
replaces #255 if successful

motivation: the execution reference is a pointer what the toolbox needs to execute. classname + method is only one type - with MCP it could also be a server + toolname.

### DIFF
--- a/src/Chain/ToolBox/ExecutionReference.php
+++ b/src/Chain/ToolBox/ExecutionReference.php
@@ -1,0 +1,14 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpLlm\LlmChain\Chain\ToolBox;
+
+final class ExecutionReference
+{
+    public function __construct(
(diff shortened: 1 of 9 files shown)


src/Chain/ToolBox/ExecutionReference.php
```diff
@@ -0,0 +1,14 @@
...
+
+namespace PhpLlm\LlmChain\Chain\ToolBox;
+
+final class ExecutionReference
+{
+    public function __construct(
+        public string $class,
+        public string $method = '__invoke',
+    ) {
+    }
```
OskarStark, line 13: we could assert, that class and method are no empty strings here

OskarStark: or use `non-empty-string` PHPDoc annotation

PR #262, opened 2025-03-19
docs: add section about different tool metadata factories

chr-hertel:


README.md
```diff
@@ -239,6 +239,42 @@ See attribute class [With](src/Chain/JsonSchema/Attribute/With.php) for all avai
...
+use PhpLlm\LlmChain\Chain\Toolbox\MetadataFactory\ReflectionFactory;
+
+$reflectionFactory = new ReflectionFactory(); // Register tools with #[AsTool] attribute
+$metadataFactory = (new MemoryFactory())      // Register or overwrite tools explicitly
+    ->addTool(...);
+$toolbox = new Toolbox(new ChainFactory($metadataFactory, $reflectionFactory), [...]);
+```
+
+> [!NOTE]
+> The order of the factories in the `ChainFactory` matters, as the first factory has the highest priority.
```
OskarStark, line 276: This is confusing, the first had the highest priority, does it really not overwrite a tool from the following then? If yes, is this really the priority we want or should the last always overwrite the first ones?

chr-hertel: It's basically like the UrlMatcher of Symfony Router. First match wins.
If you feel like that's counter intuitive we could  just flip that - I don't mind

OskarStark: Ah its about matching, fine then 👍

PR #269, opened 2025-03-29
Fix typo

Iyadhfaleh:
Hello ,

This PR is a suggestion to change the name of embedding providers variables and classes . 

If you accept my suggestion , the PR to adopt this into the bundle is almost ready .


README.md
```diff
@@ -350,23 +350,23 @@ $eventDispatcher->addListener(ToolCallsExecuted::class, function (ToolCallsExecu
 LLM Chain supports document embedding and similarity search using vector stores like ChromaDB, Azure AI Search, MongoDB
 Atlas Search, or Pinecone.
 
-For populating a vector store, LLM Chain provides the service `DocumentEmbedder`, which requires an instance of an
+For populating a vector store, LLM Chain provides the service `EmbeddingProvider`, which requires an instance of an
```
chr-hertel, line 353: this is the actual typo, right?

```suggestion
For populating a vector store, LLM Chain provides the service `Embedder`, which requires an instance of an
```

The renaming from `DocumentEmbedder` to `Embedder` was intended and part of #140

chr-hertel (changes requested):
please decouple fixing the readme (valid and good catch) and renaming the `Embedder` (not wanted)

chr-hertel (commented):
Thanks, good catch! :+1:

PR #270, opened 2025-04-08
feat: Add Raw Response Metadata Handling - GPT Token Usage Example

DZunke:
Currently it is not really possible to get into details of the interaction between user and model because the HTTPResponse is hidden behind the response converter and is then forgotten forever. So the only possibility currently to get details from the interaction would be a custom response and a custom response converter. 

As a draft and to collect your ideas and other input i want to throw this stone into the lake with having the HTTPResponse also transported into the direction of the output processors. Could this give a bit more flexibility? What do you think? 

Please have in mind this is only a draft to discuss. I thought about having this also possible if a request fails with an error status code in the future so the output processors could handle in a separation of convern way if there should be an exception thrown. Or do we maybe need an additional layer on top of the response converters itself? 

For sure this is the simplest use case with the text response in mind. Maybe for the streamed response it would need a more "recording" approach with an output processor. But would surely be possible.

### DIFF
--- a/src/Bridge/OpenAI/TokenOutputProcessor.php
+++ b/src/Bridge/OpenAI/TokenOutputProcessor.php
@@ -1,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpLlm\LlmChain\Bridge\OpenAI;
+
+use PhpLlm\LlmChain\Chain\Output;
+use PhpLlm\LlmChain\Chain\OutputProcessor;
+use PhpLlm\LlmChain\Model\Response\StreamResponse;
+
+final class TokenOutputProcessor implements OutputProcessor
+{
+    public function processOutput(Output $output): void
+    {
+        if ($output->response instanceof StreamResponse) {
+            // Streams have to be handled manually as the tokens are part of the streamed chunks
+            return;
+        }
+
+        $rawResponse = $output->response->getRawResponse();
+        if (null === $rawResponse) {
+            return;
+        }
+
+        $metadata = $output->response->getMetadata();
+
+        $metadata->add(
+            'remaining_tokens',
+            (int) $rawResponse->getHeaders(false)['x-ratelimit-remaining-tokens'][0],
+        );
+
+        $content = $rawResponse->toArray(false);
+
+        if (!\array_key_exists('usage', $content)) {
+            return;
+        }
+
+        $metadata->add('prompt_tokens', $content['usage']['prompt_tokens'] ?? null);
+        $metadata->add('completion_tokens', $content['usage']['completion_tokens'] ?? null);
+        $metadata->add('total_tokens', $content['usage']['total_tokens'] ?? null);
+    }
+}
--- a/src/Model/Response/AsyncResponse.php
+++ b/src/Model/Response/AsyncResponse.php
@@ -4,11 +4,15 @@ declare(strict_types=1);
 
 namespace PhpLlm\LlmChain\Model\Response;
 
+use PhpLlm\LlmChain\Model\Response\Exception\RawResponseAlreadySet;
+use PhpLlm\LlmChain\Model\Response\Metadata\MetadataAwareTrait;
 use PhpLlm\LlmChain\Platform\ResponseConverter;
 use Symfony\Contracts\HttpClient\ResponseInterface as HttpResponse;
 
 final class AsyncResponse implements ResponseInterface
 {
+    use MetadataAwareTrait;
+
     private bool $isConverted = false;
     private ResponseInterface $convertedResponse;
 
@@ -27,10 +31,27 @@ final class AsyncResponse implements ResponseInterface
         return $this->unwrap()->getContent();
     }
 
+    public function getRawResponse(): HttpResponse
+    {
+        return $this->response;
+    }
+
+    public function setRawResponse(HttpResponse $rawResponse): void
+    {
+        // Empty by design as the raw response is already set in the constructor and must only be set once
+        throw new RawResponseAlreadySet();
+    }
+
     public function unwrap(): ResponseInterface
     {
         if (!$this->isConverted) {
             $this->convertedResponse = $this->responseConverter->convert($this->response, $this->options);
+
+            if (null === $this->convertedResponse->getRawResponse()) {
+                // Fallback to set the raw response when it was not handled by the response converter itself
+                $this->convertedResponse->setRawResponse($this->response);
+            }
+
             $this->isConverted = true;
         }
 
--- a/src/Model/Response/Metadata/Metadata.php
+++ b/src/Model/Response/Metadata/Metadata.php
@@ -1,0 +1,99 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpLlm\LlmChain\Model\Response\Metadata;
+
+/**
+ * @implements \IteratorAggregate<string, mixed>
+ * @implements \ArrayAccess<string, mixed>
+ */
+class Metadata implements \JsonSerializable, \Countable, \IteratorAggregate, \ArrayAccess
+{
+    /**
+     * @var array<string, mixed>
+     */
+    private array $metadata = [];
+
+    /**
+     * @param array<string, mixed> $metadata
+     */
+    public function __construct(array $metadata = [])
+    {
+        $this->set($metadata);
+    }
+
+    /**
+     * @return array<string, mixed>
+     */
+    public function all(): array
+    {
+        return $this->metadata;
+    }
+
+    /**
+     * @param array<string, mixed> $metadata
+     */
+    public function set(array $metadata): void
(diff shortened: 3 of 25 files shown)


OskarStark:
Looks great, I would only vote vor having a `Metadata` object instead of an array

DZunke:
> Looks great, I would only vote vor having a `Metadata` object instead of an array

Thanks ... 👍🏼 ... i thought about extending the collection, if it is generally fine, with a key value (as object) storage and search by name, etc. so a bit of comfort in the metadata then. Just was unsure if this kind of "transport" is in the interest of the library or if someone has other restructurings in mind that would make it "easier" or more complicated.

chr-hertel:
Yeees, we need that - thanks for kicking it off! :heart_eyes:
I also like the two levels here with a more structured metadata object and the raw response.

And I would even prefer to have them lower. Basically in the `ResponseInterface` directly:

```php
namespace PhpLlm\LlmChain\Model\Response;

use PhpLlm\LlmChain\Model\Response\Metadata;
use Symfony\Contracts\HttpClient\ResponseInterface as SymfonyHttpResponse;

interface ResponseInterface
{
    /**
     * @return string|iterable<mixed>|object|null
     */
    public function getContent(): string|iterable|object|null;

    /**
     * Returns a key-value set based on provider specific response meta data.
     */
    public function getMetadata(): Metadata;

    /**
     * Return the original API response.
     */
    public function getRawResponse(): SymfonyHttpResponse;
}
```

I know that this would shift the responsibility from the `OutputProcessor` to every `ResponseConverter`, but maybe we can shift that into more clever abstractions/helper/trait/younameit if there are redundancies.

tryvin:
@chr-hertel Yes, it does! Thanks for letting me know!

I've been so busy dealing with external providers that I didn't have the time to do what I promised in the issue.

But I will keep an eye on this PR

DZunke:
> Yeees, we need that - thanks for kicking it off! 😍 I also like the two levels here with a more structured metadata object and the raw response.
> 
> And I would even prefer to have them lower. Basically in the `ResponseInterface` directly:
> 
> ```
> namespace PhpLlm\LlmChain\Model\Response;
> 
> use PhpLlm\LlmChain\Model\Response\Metadata;
> use Symfony\Contracts\HttpClient\ResponseInterface as SymfonyHttpResponse;
> 
> interface ResponseInterface
> {
>     /**
>      * @return string|iterable<mixed>|object|null
>      */
>     public function getContent(): string|iterable|object|null;
> 
>     /**
>      * Returns a key-value set based on provider specific response meta data.
>      */
>     public function getMetadata(): Metadata;
> 
>     /**
>      * Return the original API response.
>      */
>     public function getRawResponse(): SymfonyHttpResponse;
> }
> ```
> 
> I know that this would shift the responsibility from the `OutputProcessor` to every `ResponseConverter`, but maybe we can shift that into more clever abstractions/helper/trait/younameit if there are redundancies.
> 
> @4lxndr @tryvin that would also help with your use-cases, right?

chr-hertel:
While working on #280 and with that fake PipelineResponse I was wondering if we should rather use terminology like Input/Output on this layer already/as well - instead of Request/Response - what do you think?

DZunke:
Would this not very confusing? Then there would be two different "Output" named layers within the data flow.

chr-hertel:
Yeah true, having that terminology twice could be annoying 🤔

OskarStark:
ModelResponse? ModelOutput? To be more specific regarding the layer? Just thinking out loud

src/Bridge/OpenAI/TokenOutputProcessor.php
```diff
@@ -0,0 +1,37 @@
...
+        $content = $rawResponse->getContent(false);
+        $content = \json_decode($content, true);
+
+        if (!\array_key_exists('usage', $content)) {
+            return;
+        }
+
+        $metadata->add('prompt_tokens', $content['usage']['prompt_tokens'] ?? null);
+        $metadata->add('completion_tokens', $content['usage']['completion_tokens'] ?? null);
+        $metadata->add('total_tokens', $content['usage']['total_tokens'] ?? null);
```
OskarStark, line 35: I think it could be handy to be able to provide an array

```php
$metadata->add([
    'prompt_tokens' => ......,
]);
```

chr-hertel:
Nevermind the pipeline if it's only the commit thingy - need to dig deeper why that is so flaky

src/Model/Response/Metadata/EmptyMetadataTrait.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpLlm\LlmChain\Model\Response\Metadata;
+
+trait EmptyMetadataTrait
```
chr-hertel, line 7: to me it would be more consistent to be also `MetadataAwareTrait` - sounds a bit like metadata would be empty when using this one, but in reality it's just initializing that

chr-hertel:
from an OOP point of view - why choose traits over an abstract base implementation of the response? isn't it that all response classes have the two traits?

with having the interface people could still decide to do their own - or we could do both - having a base class that uses both traits. not sure if i miss the downside here

src/Bridge/OpenAI/TokenOutputProcessor.php
```diff
@@ -0,0 +1,43 @@
...
+
+        $metadata = $output->response->getMetadata();
+
+        $metadata->add(
+            'remaining_tokens',
+            (int) $rawResponse->getHeaders(false)['x-ratelimit-remaining-tokens'][0],
+        );
+
+        $content = $rawResponse->getContent(false);
+        $content = \json_decode($content, true);
```
chr-hertel, line 33:
```suggestion
        $content = $rawResponse->toArray(false);
```

src/Model/Response/AsyncResponse.php
```diff
@@ -27,10 +31,27 @@ public function getContent(): string|iterable|object|null
...
+    }
+
     public function unwrap(): ResponseInterface
     {
         if (!$this->isConverted) {
             $this->convertedResponse = $this->responseConverter->convert($this->response, $this->options);
+
+            if (null === $this->convertedResponse->getRawResponse()) {
+                // Fallback to set the raw response when it was not handled by the response converter itself
+                $this->convertedResponse->setRawResponse($this->response);
```
chr-hertel, line 52: this is currently the path for the most cases, right?

src/Model/Response/Metadata/Metadata.php
chr-hertel, line 1: smooth :) :+1:

src/Model/Response/AsyncResponse.php
DZunke: Yeah. I just wanted to open up that people can do it also in the response converter for themself ... you never know.

src/Model/Response/Metadata/EmptyMetadataTrait.php
DZunke: Thanks! You are right, i rename it 👍🏼

DZunke:
> from an OOP point of view - why choose traits over an abstract base implementation of the response? isn't it that all response classes have the two traits?
> 
> with having the interface people could still decide to do their own - or we could do both - having a base class that uses both traits. not sure if i miss the downside here

Mhm. Yeah. First i wanted to have a most flexible approach but after revisiting it i have implemented an abstract base response to extend from if needed. But i also kept the traits as a handy helpers, if this is fine? See `AsyncResponse` where i have not extended from the `BaseResponse` but utilized the `MetadataAwareTrait` as it has it's own raw symfony response usage. 

What do you think?

src/Bridge/OpenAI/TokenOutputProcessor.php
OskarStark: what about `function add(string|array $value, mixed $content)` while content should be ignored then ofc.
I don't like the fuzzy syntax, but have no other ideas, maybe let's skip the idea completely for now @chr-hertel ?

chr-hertel: really two sided here - i agree that having both in one methods feels a bit overloaded, but also see the argument for the API usage being redundant here.

what do you think of using Symfony's `ParameterBag` as a blueprint here?
https://github.com/symfony/http-foundation/blob/7.2/ParameterBag.php#L60-L81

Anyhow, not a blocker for this PR and could easily be a follow up - i think the most important part here is the introduction :muscle:

src/Model/Response/AsyncResponse.php
chr-hertel: yup, fine, was just curious if i read it right

chr-hertel (approved):
Thanks a ton for taking care about this @DZunke - that feature was overdue!

OskarStark:
Nice one 👏

PR #276, opened 2025-04-12
feat: bring in Llama 3.3 support with Azure

chr-hertel:


chr-hertel:
this should later be more and adapter of Llama using GPT contract on Azure.

PR #278, opened 2025-04-13
refactor: rename Model::version to Model::name

chr-hertel:
I'm currently working on an integration for Hugging Face and I think `$version` is misleading since this could also use quite different Models of one family.

Looking more into the direction of `GenericModel` what we saw already with #203 
https://github.com/php-llm/llm-chain/blob/main/src/Bridge/OpenRouter/GenericModel.php

WDYT @OskarStark @DZunke

### DIFF
--- a/examples/chat-claude-anthropic.php
+++ b/examples/chat-claude-anthropic.php
@@ -16,7 +16,7 @@ if (empty($_ENV['ANTHROPIC_API_KEY'])) {
 }
 
 $platform = PlatformFactory::create($_ENV['ANTHROPIC_API_KEY']);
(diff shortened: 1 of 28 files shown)


OskarStark (approved):

DZunke:
Just because i was pinged. I think the `name` wording is better in the domain model. The naming (i checked google, anthropic and open ai) for the models is "model" and so within the Model it is not the version but the name of the model. 

For me this fits and should make more clear what is meant. So: 👍 - Thanks, also for the info.

chr-hertel:
Thanks @DZunke for the feedback :)

PR #280, opened 2025-04-14
feat: add support for TransformersPHP for model runtime in PHP

chr-hertel:
Sitting on top of https://github.com/CodeWithKyrian/transformers-php


src/Bridge/TransformersPHP/PipelineResponse.php
chr-hertel, line 1: ouchy

PR #281, opened 2025-04-14
ci: skip mongodb extension for now

chr-hertel:
For now the easiest solution as the pipeline doesn't need the extension anyways.


chr-hertel:
@OskarStark If you have an idea how to deal with the MongoDB 2.0 thingy, feel free to jump into this :)

OskarStark:
Imho we can just allow the 2.0 additionally, but I need to try it out. Can you please create an issue? Thanks

PR #285, opened 2025-04-18
refactor: message content file handling [BC BREAK]

chr-hertel:
Preparation for #279 

* Introducing a base File that centralizes mime type, base64 handling, etc.
* Image split into ImageUrl and Image based on File => i usually know what i need in user land
  * `new Image('https://...')` => `new ImageUrl('https://...)`
  * `new Image('data:image/jpeg;base64,...')` => `Image::fromDataUrl('data:image/jpeg;base64,...')`
  * `new Image('/path/to/jpeg.jpg')` => `Image::fromFile('/path/to/jpeg.jpg')`
* removing `File` class of Whisper bridge in favor of `Audio`


tests/Model/Message/Content/ImageTest.php
```diff
@@ -2,55 +2,37 @@
 
 declare(strict_types=1);
 
-namespace PhpLlm\LlmChain\Tests\Model\Message\Content;
+namespace Model\Message\Content;
```
OskarStark, line 5: Namespace looks wrong

chr-hertel: solved with #289 - thanks!

PR #288, opened 2025-04-21
fix: JsonSchema required

lvluoyue:


chr-hertel:
Hi @lvluoyue, thanks for taking care of that!

How did you run into it?

I was wondering before if that's needed or only with stricr mode

lvluoyue:
> Hi @lvluoyue, thanks for taking care of that!
> 
> How did you run into it?
> 
> I was wondering before if that's needed or only with stricr mode

Parameters should be optional when specifying a default value

```php
    public function searchSong(
        string $keyword,
        #[With(minimum: 1, maximum: 60)]
        int    $limit = 10,
        #[With(minimum: 1)]
        int    $page = 1
    ): string {
        return R::mcp($this->tencentService->searchSong($keyword, $page, $limit));
    }
```

chr-hertel:
Sounds good, yes!

Do you mind sharing what's that `R::MCP(...)` - something custom or a library?

lvluoyue:
> 听起来不错，是的！
> 
> 你介意分享那是什么 - 自定义的东西还是库？`R::MCP(...)`

The full name of R is response, and it is used to encapsulate various response formats into a unified format

chr-hertel:
you can run the failing tests locally with:
```
php vendor/bin/phpunit tests/Chain/JsonSchema/FactoryTest.php tests/Chain/Toolbox/ToolboxTest.php
```

OskarStark:
Maybe I get it wrong but in your example the parameter is required but with a default value, so null is not supported here, or am I missing sth?

OskarStark:
Are you using OpenAI?

OskarStark:
https://platform.openai.com/docs/guides/function-calling#strict-mode

chr-hertel:
if you revert commit b6007f4bd1d05a6bbc2c4b866995e0f51226429d the pipeline should be fine - i personally don't see real value in those `@throws` tags on method tbh, besides interface docs

chr-hertel (approved):
Thanks @lvluoyue! Looks good to me now 👍

PR #290, opened 2025-04-22
feat: adding basic Brave search tool and crawler

chr-hertel:


README.md
```diff
@@ -340,7 +340,9 @@ $eventDispatcher->addListener(ToolCallsExecuted::class, function (ToolCallsExecu
 
 #### Code Examples (with built-in tools)
 
+1. **Brave Tool**: [toolbox-brave.php](examples/toolbox-brave.php)
 1. **Clock Tool**: [toolbox-clock.php](examples/toolbox-clock.php)
+1. **Crawler Tool**: [toolbox-brave.php](examples/toolbox-brave.php)
```
OskarStark, line 345: Same file as above?

src/Chain/Toolbox/Tool/Crawler.php
```diff
@@ -0,0 +1,31 @@
...
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+#[AsTool('crawler', 'A tool that crawls one page of a website and returns the visible text of it.')]
+final readonly class Crawler
+{
+    public function __construct(
+        private HttpClientInterface $httpClient,
+    ) {
+        if (!class_exists(DomCrawler::class)) {
+            throw new \RuntimeException('The DomCrawler component is not installed. Please install it using "composer require symfony/dom-crawler".');
```
OskarStark, line 18: If we have an own runtime exception we should use it. Cannot check, I am currently on a phone 📱

chr-hertel: oh the exception usage is quite inconsistent - will tackle in a follow-up and streamline all places :+1:

PR #291, opened 2025-04-22
feat: add support for mistral

chr-hertel:
* [x] text generation
* [ ] streaming
* [ ] tool calling
* [ ] image input
* [ ] embeddings
* [ ] examples
* [ ] documentation


src/Bridge/Mistral/ResponseConverter.php
chr-hertel, line 1: Sooo much duplication with the GPT one ...

src/Chain/Toolbox/Metadata.php
```diff
@@ -41,6 +41,9 @@ public function jsonSerialize(): array
 
         if (isset($this->parameters)) {
             $function['parameters'] = $this->parameters;
+        } else {
+            // ⚠ Necessary for Mistral, but breaks OpenAI ⚠
+            $function['parameters'] = ['type' => 'object'];
```
chr-hertel, line 46: this needs a better solution

chr-hertel: test failure is real

PR #296, opened 2025-05-04
chore: renamen model toolbox examples to toolcall

chr-hertel:


OskarStark:
No change in the README ?

chr-hertel:
Thanks @OskarStark, fixed with #300

PR #297, opened 2025-05-04
feat: add image input support for Claude

chr-hertel:
This would also break GPT & Mistral implementation


chr-hertel:
closed in favor of #298

PR #300, opened 2025-05-04
docs: fix after example rename in #296

chr-hertel:


examples/openai/toolcall.php
chr-hertel, line 1: we had two streaming examples for openai toolcalls, whoops

PR #301, opened 2025-05-04
refactor: switch from `JsonSerializable` to normalizer

chr-hertel:
```
`8.`888b                 ,8'  8 8888 8 888888888o   
 `8.`888b               ,8'   8 8888 8 8888    `88. 
  `8.`888b             ,8'    8 8888 8 8888     `88 
   `8.`888b     .b    ,8'     8 8888 8 8888     ,88 
    `8.`888b    88b  ,8'      8 8888 8 8888.   ,88' 
     `8.`888b .`888b,8'       8 8888 8 888888888P'  
      `8.`888b8.`8888'        8 8888 8 8888         
       `8.`888`8.`88'         8 8888 8 8888         
        `8.`8' `8,`'          8 8888 8 8888         
         `8.`   `8'           8 8888 8 8888         
```

Trying to address pain of #291 and #297

- [ ] fix bedrock examples ...


src/Bridge/OpenAI/DallE/ContractExtension.php
chr-hertel, line 1: this shouldn't be needed - would be easier if we do it directly in ModelClient, like before

src/Bridge/OpenAI/Embeddings/ContractExtension.php
chr-hertel, line 1: same here

src/Platform/ResponseConverter.php
```diff
@@ -13,7 +13,7 @@ interface ResponseConverter
     /**
      * @param array<string, mixed>|string|object $input
      */
-    public function supports(Model $model, array|string|object $input): bool;
+    public function supports(Model $model): bool;
```
chr-hertel, line 16: separate PR

src/Platform/Contract.php
chr-hertel, line 1: questionable instance handling, would be great tho to decouple from globale `NormalizerInterface`

src/Bridge/OpenAI/DallE/ContractExtension.php
```diff
@@ -0,0 +1,31 @@
...
+    {
+        return $model instanceof DallE;
+    }
+
+    public function registerTypes(): array
+    {
+        return [
+            'string' => 'handleInput',
+        ];
+    }
```
OskarStark, line 23: couldn't we do this using an attribute like the event listeners do?

src/Bridge/OpenAI/DallE/ContractExtension.php
```diff
@@ -0,0 +1,31 @@
...
+
+    public function handleInput(string $input): array
```
OskarStark, line 25:
```suggestion
    #[AsExtension('string')]
    public function handleInput(string $input): array
```
idea

src/Bridge/OpenAI/DallE/ContractExtension.php
chr-hertel: reverted back to plain Normalizer since i had to start doing recursive object traversing and array handling myself. not worth it at the moment.

chr-hertel:
closed in favor of #326

PR #305, opened 2025-05-12
refactor: introduce model capabilities with generic base model instead of interfaces with supportsX methods

chr-hertel:
Idea is to make it possible to use models from more generic model providers like huggingface, replicate, and more with chain as well and configure the capabilities via provided classes instead of interfaces.

besides having a generic model that you can basically configure like you want, it introduces a new way of defining specific models in a bridge, e.g. 
```php
class DallE extends Model
{
    public const DALL_E_2 = 'dall-e-2';
    public const DALL_E_3 = 'dall-e-3';

    /** @param array<string, mixed> $options The default options for the model usage */
    public function __construct(string $name = self::DALL_E_2, array $options = [])
    {
        $capabilities = [
            Capability::INPUT_TEXT,
            Capability::OUTPUT_IMAGE,
        ];

        parent::__construct($name, $capabilities, $options);
    }
}
```


chr-hertel:
one question would be if we want to differentiate between three kinds of capabilities:
* input capabilities, e.g. supports text or audio
* output capabilities, e.g. support images
* functional capabilities, e.g. supports tool calling

:thinking:

chr-hertel:
closed in favor of #326

PR #310, opened 2025-05-16
fix: missing namespace when upserting in Pinecone store

sarjon:
When adding documents, it does not take configured namespace into consideration, this PR fixes it.


chr-hertel (approved):
That slipped through - thanks for catching and fixing it @sarjon!

PR #312, opened 2025-05-17
feat: implement Bedrock and model clients for Claude, Llama, and Nova

bjalt:
- [x] Test for Prompt Converters
- [x] Documentation for limits of invokeModel with tool calling using Nova models
- [x] ~~Documentation for configuration~~ should be part of the bundle
- [x] Image support for Claude and Nova using Bedrock
- [x] Remove local model validation
- [x] Example usage for Bedrock


src/Bridge/Bedrock/Meta/LlamaModelClient.php
```diff
@@ -0,0 +1,75 @@
...
+        if (!isset($responseBody['generation'])) {
+            throw new \RuntimeException('Response does not contain any content');
+        }
+
+        return new TextResponse($responseBody['generation']);
+    }
+
+    private function getModelId(Model $model): string
+    {
+        $euRegions = ['eu-west-1', 'eu-central-1', 'eu-west-2', 'eu-west-3', 'eu-south-1', 'eu-north-1'];
```
OskarStark, line 60: I am wondering if we should do this validation on the client side. What does the api return if you request an unsupported region? 

Keep in mind that we would need a new release whenever a new region is added

bjalt: That's a good idea. The API will return a 400 with something along the lines of "Invalid model identifier".

bjalt: Remove the region listing.

bjalt:
There is some code duplication in https://github.com/dekkode/llm-chain/blob/bedrock/src/Bridge/Anthropic/ModelHandler.php#L74-L122 and https://github.com/dekkode/llm-chain/blob/bedrock/src/Bridge/Bedrock/Anthropic/ClaudeHandler.php#L69-L117.
For the sake of keeping this PR small I left it in.

bjalt:
Fixes #312

chr-hertel:
First of all thanks for working on this - I think it's a great addition and a very valid case in challenging the current extension points of the Platform layer - good timing for that large refactoring in #316 - but i guess we should get this merged before the other one.

also the redundancy can be tackled as a follow up - that's currently a matter how the response structured is handled.

please have a look at the phpstan failure, but nevermind the commit message check - just a reminder for me while merging.

i will give it a detailed look right away - thanks again! :+1:

src/Bridge/Bedrock/Nova/Nova.php
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    public function supportsAudioInput(): bool
+    {
+        return false;
+    }
+
+    public function supportsImageInput(): bool
+    {
+        return true;
```
chr-hertel, line 42: only from the docs i'd say it should exclude the micro model, right?

chr-hertel:
Don't have access to the models on AWS atm, but will trust on your tests here - can you please add some examples though?

bjalt:
@chr-hertel 
> please have a look at the phpstan failure, but nevermind the commit message check - just a reminder for me while merging.

I just looked at the phpstan issue and they should not be there. It runs locally and the code is guarded with an if statement limiting the possible type that can be handled.
Do you utilize any kind of caching in the actions? Can you trigger the qa run again before I set up the GitHub action in my repo to reproduce the issue?

src/Bridge/Bedrock/Nova/Nova.php
bjalt: Done

bjalt:
@chr-hertel I fixed the ci issues and added some examples.

examples/bedrock/chat-nova-binary.php
chr-hertel, line 1: That `-binary` suffix might be copy/paste error - at least i don't spot files here

chr-hertel:
needs a rebase, but other than that i'd like to merge this - thanks already! :)

chr-hertel (approved):
Thanks for you work @bjalt - will continue to get rid off redundencies :+1:

PR #320, opened 2025-05-19
WIP: AWS Bedrock Support, Google Gemini Function calling, and file support, and more.

tryvin:
- AWS Bedrock
- Google Gemini function calling, file support, and more
- CerebrasAI support
- Basis for OpenAI-compatible API generic providers

## Explanation

Hey guys, I'm finally adding what I have been working on in other projects and bringing it up-to-date with the new Metadata and Output processor for token counts.

This also adds new providers, and an "OpenAI generic" provider is also coming.

I created the PR so you are aware of what I'm doing and prevent double work, and of course, so you can start the review while I finish the tests and more.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -56,7 +56,8 @@
         "symfony/event-dispatcher": "^6.4 || ^7.1",
         "symfony/finder": "^6.4 || ^7.1",
         "symfony/process": "^6.4 || ^7.1",
-        "symfony/var-dumper": "^6.4 || ^7.1"
+        "symfony/var-dumper": "^6.4 || ^7.1",
+        "aws/aws-sdk-php": "^3"
     },
     "suggest": {
         "codewithkyrian/chromadb-php": "For using the ChromaDB as retrieval vector store.",
@@ -64,7 +65,8 @@
         "mongodb/mongodb": "For using MongoDB Atlas as retrieval vector store.",
         "probots-io/pinecone-php": "For using the Pinecone as retrieval vector store.",
         "symfony/css-selector": "For using the YouTube transcription tool.",
-        "symfony/dom-crawler": "For using the YouTube transcription tool."
+        "symfony/dom-crawler": "For using the YouTube transcription tool.",
+        "aws/aws-sdk-php": "For using AWS Bedrock"
     },
     "config": {
         "allow-plugins": {
--- a/examples/google/toolcall.php
+++ b/examples/google/toolcall.php
@@ -1,0 +1,33 @@
+<?php
+
+use PhpLlm\LlmChain\Bridge\Google\Gemini;
+use PhpLlm\LlmChain\Bridge\Google\PlatformFactory;
+use PhpLlm\LlmChain\Chain;
+use PhpLlm\LlmChain\Chain\Toolbox\ChainProcessor;
+use PhpLlm\LlmChain\Chain\Toolbox\Tool\YouTubeTranscriber;
+use PhpLlm\LlmChain\Chain\Toolbox\Toolbox;
+use PhpLlm\LlmChain\Model\Message\Message;
+use PhpLlm\LlmChain\Model\Message\MessageBag;
+use Symfony\Component\Dotenv\Dotenv;
+use Symfony\Component\HttpClient\HttpClient;
+
+require_once dirname(__DIR__, 2).'/vendor/autoload.php';
+(new Dotenv())->loadEnv(dirname(__DIR__, 2).'/.env');
+
+if (empty($_ENV['GOOGLE_API_KEY'])) {
+    echo 'Please set the GOOGLE_API_KEY environment variable.'.PHP_EOL;
+    exit(1);
+}
+
+$platform = PlatformFactory::create($_ENV['GOOGLE_API_KEY']);
+$llm = new Gemini(Gemini::GEMINI_2_FLASH);
+
+$transcriber = new YouTubeTranscriber(HttpClient::create());
+$toolbox = Toolbox::create($transcriber);
+$processor = new ChainProcessor($toolbox);
+$chain = new Chain($platform, $llm, [$processor], [$processor]);
+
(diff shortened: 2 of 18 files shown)


chr-hertel:
Oh wow, that's a ton - thanks for bringing that forward! :pray: 

I think three things would be important here:
* Let's try to extract some concerns from this PR to get it sorted easier, e.g. extended Gemini support or events
* Can you take a look at these PR and tell which should go first or wait?
  * #305 
  * #301 
  I could imagine that this might help with stuff you needed to sort here
* And of course, what's your take on #312?

Cheers

tryvin:
Hey @chr-hertel, yeah, these are things I had to do in my projects that got scattered all over the place, so I'm integrating all of them into the architecture of this project.

Now, answering your questions:

- Let's try to extract some concerns from this PR to get it sorted more easily, e.g., extended Gemini support or events
   - Agreed, lemme try to finish extracting everything I had into this big PR, and then we can decide on what should be extracted and what shouldn't. There's still work to do here.

- Can you take a look at these PR and tell which should go first or wait? https://github.com/php-llm/llm-chain/pull/305 https://github.com/php-llm/llm-chain/pull/301 I could imagine that this might help with stuff you needed to sort here
  - I would say 301 should go first, it's a lot of refactor,ing but it will help a lot with adding more providers and making things simple. Sadly, not every provider has the same input and output across the board.
 
- And of course, what's your take on https://github.com/php-llm/llm-chain/pull/312?
   - We might want to merge things at some point here, I noticed he is using a wrapper for bedrock, in my first iteration (in one project here), I used AWS SDK directly and wrapped the response in `HttpResponseInterface` implementation, but this time I implemented it directly via REST calls, and only used the AWS SDK to sign the headers so I don't need to care about how to get the credentials and sign it correctly.

tryvin:
@chr-hertel I think it's now stable enough for me to start writing tests for it, and for us to decide how I should split this PR.

Thoughts?

chr-hertel:
Hey, i hope i got it right, but i'd propose to split it into 4 different PRs:

1. Event Dispatching - was wondering tho why in the Toolbox' processor or if it belongs directly in the chain
2. Gemini Improvements
3. Cerebras Implementation

and of course

4. AWS Bedrock Implementation - but here I'd need another thought what the better approach is - compared to #312 - I personally use AWS too little to have an opinion atm

so, best to start with 1-3? wasn't able to progress with #301 this week - and weekend looks busy as well

tryvin:
@chr-hertel
- Event Dispatching - was wondering tho why in the Toolbox' processor or if it belongs directly in the chain 

So, the idea is that in order to keep the history correct and track all tool calls and usage tokens, I need access to the intermediary messages, and the way the chain was it was hiding all these intermediary messages and only calling the output processor at the end with the last response that wasn't a tool call. 

We can maybe change it to some sort of "Intermediary Processor" 

- AWS Bedrock Implementation - but here I'd need another thought what the better approach is

We are achieving similar results, I focused initially on Nova and Titan embeddings models, and used the raw Http client. Both of us are using external packages though, I need aws SDK for the credentials signature, and the other PR is using an external implementation for calling the service itself, and transforming the response (kinda of what I did initially the first time I interacted with bedrock and php-llm)

IMHO, it's better to use the official SDK and rely only on Http requests so you can keep it "uniform", but that's just my thought.

chr-hertel:
> We can maybe change it to some "Intermediary Processor"

uh, interesting - i like that :+1:

> it's better to use the official SDK and rely only on HTTP requests so you can keep it "uniform"

I def agree on that uniform argument, but another thought came to my mind - maybe you can confirm or challenge that:

It is quite likely that LLM Chain gets integrated into an existing application, that also has already it's AWS infrastructure layer. And with that infrastructure layer it was already decided to use the SDK or the Async lib. 
I'd say it would actually be a nice feature for LLM Chain not to mess with that decision but basically offer both - `PlatformFactory::createAsync` and `PlatformFactory::createSdk` (or better naming :D) - and model classes could actually be the same i guess. is there a logic within that thought? WDYT? cc @bjalt

tryvin:
@chr-hertel The thing is, the SDK just allows you to invoke the model, AWS calls then "Foundation Models", so everything else is on you.

Tool calling management, message bags, and everything.

I mean, I only used the SDK because I didn't want to handle AWS signatures

chr-hertel:
gave it another thought and would go with both AWS implementations - one relies on that async package and one the sdk - like you did. still makes sense to me - would need docs and reasonable naming tho

valtzu:
Thanks for the great work @tryvin! I can't wait for Google Gemini tool support, so I have extracted it into a separate PR #331, I hope you don't mind. Hopefully this will get it merged sooner.

tryvin:
@chr-hertel, I suppose with the latest changes merged to master, I will need to merge them into my working branch and recreate some of that code.

However, to prioritize things, what do you think the order would be?

- Improve the AWS Bedrock Bridge to support both AsyncAWS and Official AWS PHP-SDK
- OpenAI Generic Bridge (for providers that support an OpenAI-like RESTful API)
- CerebrasAI "Official API" Support (usually used by enterprises as they use Hugging Face and OpenRouter for "consumer grade")

Thanks!

chr-hertel:
Hey @tryvin, it's open source, your freetime, so could go with whatever is most fun for you ---- buuuut, since you're asking 😂 

I'd go with this:
# **OpenAI Generic Bridge**
  I'd like to understand what still is needed to skip that generic bridge but just use the `Model` class and the default contract, that should be OpenAI compatible - at least that was the idea - maybe I forgot some thing, so you could test that and prove me wrong
# **Extending AWS Bedrock Bridge**
  I'm still not super happy with the Platform layer and the extension points, so additional provider/models/use-cases help with that - and I'd prioritize AWS over Cerebras due to popularity - maybe I'm wrong
# **CerebrasAI**
  Also quite interested to see that now, since the response converter and error handling is still not great in the platform layer and additional use-cases help to understand to best abstraction - basically like the above but less common I'd guess

Thanks for sticking around and helping to improve here - much appreciated 🙏

PR #326, opened 2025-05-29
refactor: Major Library Refactoring

chr-hertel:
- [ ] Introducing global model class to be more agnostic with explicit `Capability` class
- [ ] Introducing Contract with Normalizers to solve Base Contract (OpenAI) vs. Model/Platform specific differences
- [ ] Reshape into three major components `Chain`, `Platform` and `Store`
- [ ] Adoption Symfony-style `Interface` suffix

Replaces #301 and #305


examples/bedrock/plain-toolcall.php
chr-hertel, line 1: Hey @bjalt, struggling to get nova tool calling right, any idea?

i always get:
> Malformed input request: #/messages/2/content/0: required key [toolUse] not found, please reformat your input and try again.

PR #328, opened 2025-06-01
chore: add authors

chr-hertel:
I hope i got it more or less right based on the git history - let me know if I mixed sth up


OskarStark:
Thanks for taking care of this tedious work 🙏

PR #332, opened 2025-06-15
[DeepSeek][Tools] Allow to change URL for OpenAI  GPT using base_url configuration option

ksn135:
DeepSeek now support function calling https://api-docs.deepseek.com/guides/function_calling

```php
<?php // config/packages/lim_chain.php

use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;

return static function (ContainerConfigurator $containerConfigurator): void {
    $containerConfigurator->extension('llm_chain', [
        'platform' => [ 'openai' => [ 'api_key' => '%env(DEEPSEEK_API_KEY)%']],
        'chain' => [
            'default' => [
                'model' => [
                    'name' => 'GPT', // keep it
                    'version' => '%env(DEEPSEEK_API_MODEL)%', // 'deepseek-chat'
                    'options' => [
                        'base_url' => '%env(DEEPSEEK_API_URL)%', // 'https://api.deepseek.com'
                    ],
                ],
                'system_prompt' => 'You are a helpful assistant and you can provide the current date and time.',
            ],
        ],
    ]);
};
```

```php
<?php // src/AI/Tool/Clock.php
declare(strict_types=1);

namespace App\AI\Tool;

use PhpLlm\LlmChain\Chain\ToolBox\Attribute\AsTool;
use Symfony\Component\Clock\ClockInterface;
use Symfony\Component\Clock\NativeClock;

#[AsTool('clock', description: 'Provides the current date and time with time zone.')]
final readonly class Clock
{
    public function __construct(
        private ClockInterface $clock = new NativeClock(),
    ) {
    }

    public function __invoke(): string
    {
        return sprintf(
            'Current date is %s (YYYY-MM-DD) and the time is %s (HH:MM:SS) and time zone is %s.',
            $this->clock->now()->format('Y-m-d'),
            $this->clock->now()->format('H:i:s'),
            $this->clock->now()->getTimezone()->getName(), // for correct result needs the time zone !!!
        );
    }
}

```
### Using Tools – quite definitely!!!

Screenshot 2025-06-15 at 20 52 17

Screenshot 2025-06-15 at 20 52 04

### Without tools – hallucinating!!! 

Screenshot 2025-06-15 at 19 27 17

Screenshot 2025-06-15 at 19 26 58

### DIFF
--- a/src/Platform/Bridge/OpenAI/GPT/ModelClient.php
+++ b/src/Platform/Bridge/OpenAI/GPT/ModelClient.php
@@ -35,7 +35,7 @@ final readonly class ModelClient implements PlatformResponseFactory
         return $model instanceof GPT;
     }
 
-    public function request(Model $model, array|object|string $input, array $options = []): ResponseInterface
+    public function request(Model $model, array|string $payload, array $options = []): ResponseInterface
     {
         $base_url = $model->getOptions()['base_url'] ?? 'https://api.openai.com/v1';
         $chat_completions_url = $model->getOptions()['chat_completions_url'] ?? '/chat/completions';
@@ -45,10 +45,7 @@ final readonly class ModelClient implements PlatformResponseFactory
             "{$base_url}{$chat_completions_url}",
             [
                 'auth_bearer' => $this->apiKey,
-                'json' => array_merge($options, [
-                    'model' => $model->getVersion(),
-                    'messages' => $input,
-                ]),
+                'json' => array_merge($options, $payload),
             ]
         );
     }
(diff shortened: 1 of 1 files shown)


chr-hertel (changes requested):
Thanks for giving this a shot!
Instead of changing the base_url for deepseek.com, we should bring in a deepseek bridge.

I know that there is a lot of similarity, and we're getting rid off the redundancy step by step, but reusing the OpenAI bridge for a different model and platform is violating the idea of bridges. the contract handling should already work out of the box only the response conversion is still copy+paste.

chr-hertel:
@ksn135 if you're up for it, sure :)

ksn135:
Well, close it for now.

PR #337, opened 2025-06-22
chore: add Claude.md for Claude Code support

chr-hertel:


chr-hertel:
feels like this belongs into the global gitignore and not in the repository :thinking:

PR #340, opened 2025-06-24
Allow `\Stringable` for `Message::ofUser()` and `Message:forSystem()`

OskarStark:


src/Platform/Message/Message.php
```diff
@@ -32,10 +32,10 @@ public static function ofAssistant(?string $content = null, ?array $toolCalls =
         return new AssistantMessage($content, $toolCalls);
     }
 
-    public static function ofUser(string|ContentInterface ...$content): UserMessage
+    public static function ofUser(\Stringable|string|ContentInterface ...$content): UserMessage
     {
         $content = array_map(
-            static fn (string|ContentInterface $entry) => \is_string($entry) ? new Text($entry) : $entry,
+            static fn (string|ContentInterface $entry) => \is_string($entry) ? new Text($entry) : ($entry instanceof \Stringable ? new Text((string) $entry) : $entry),
```
valtzu, line 38: 1. I think this would not work correctly if a `ContentInterface` implementation implements `\Stringable` too – maybe it'd be better to check for `ContentInterface` instead of `string` & `\Stringable`?
2. The check for `\Stringable` is likely never hit because implicit cast is done because of `\Stringable` missing from the closure `$entry` param type

OskarStark: Makes sense, will fix it, thanks for the hint

chr-hertel (approved):
Thanks @OskarStark

Issue #28, opened 2024-09-23
Models, Platforms & Stores Support


chr-hertel:
# Models, Platforms & Stores
Potential targets for supported models, platforms and stores:

## Language Models
* [x] [GPT](https://platform.openai.com/docs/models/overview)
* [ ] [Anthropic Claude](https://www.anthropic.com/claude) (only partially done yet)
  * [ ] Tool Support (w/o streaming): #209 
* [ ] [Google Gemini](https://cloud.google.com/vertex-ai/generative-ai/docs/learn/models#gemini-models)
  * With [OpenRouter](https://openrouter.ai/) #203 
  * #215 
  * #220
* [ ] [Google Gemma](https://cloud.google.com/vertex-ai/generative-ai/docs/learn/models#gemma-models)
* [ ] [Meta Llama](https://www.llama.com/)
  * First Steps: #73
  * Azure Support: #276 
* [ ] [Mistral](https://docs.mistral.ai/getting-started/models/models_overview/)
  * Support with Mistral API with #291 
* [ ] [Microsoft Phi](https://azure.microsoft.com/en-us/products/phi)
* [ ] [DeepSeek](https://www.deepseek.com/)
  * With [OpenRouter](https://openrouter.ai/) #203 

## Embeddings Models
* [x] [OpenAI Text Embeddings](https://platform.openai.com/docs/guides/embeddings/embedding-models)
* [x] [Voyage Embeddings](https://docs.voyageai.com/docs/embeddings)
  * #64 
* [ ] [Gecko](https://cloud.google.com/vertex-ai/generative-ai/docs/learn/models#models)
* [ ] [Mistral Embed](https://docs.mistral.ai/capabilities/embeddings/)
  * Support with Mistral API with #291 
* [x] Google Embeddings
  * #347 

## Platforms
* [x] [OpenAI](https://platform.openai.com/docs/overview)
* [x] [Anthropic](https://www.anthropic.com/)
* [x] [Voyage](https://www.voyageai.com/) 
  * #64
* [ ] [Azure](https://azure.microsoft.com/en-gb/solutions/ai/) (only GPT right now)
  * Llama support with #276 
  * Whisper support with #252 
* [ ] [GitHub Marketplace](https://github.com/marketplace/models) (basically Azure?)
* [ ] [Google Cloud](https://cloud.google.com/model-garden)
* [ ] [Amazon Bedrock](https://aws.amazon.com/bedrock/)
[...]

PR #342, opened 2025-06-24
feat: add MariaDB store

valtzu:
Related to #28

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,9 +37,11 @@
         "webmozart/assert": "^1.11"
     },
     "require-dev": {
+        "ext-pdo": "*",
         "codewithkyrian/chromadb-php": "^0.2.1 || ^0.3 || ^0.4",
         "codewithkyrian/transformers": "^0.5.3",
         "async-aws/bedrock-runtime": "^0.1.0",
+        "doctrine/dbal": "^3.0 || ^4.0",
         "mongodb/mongodb": "^1.21 || ^2.0",
         "php-cs-fixer/shim": "^3.70",
         "phpstan/phpstan": "^2.0",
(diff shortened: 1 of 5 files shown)


src/Store/Bridge/MariaDB/Store.php
```diff
@@ -0,0 +1,153 @@
...
+
+    private function toBinary(Uuid $uuid): string
+    {
+        return $uuid->toBinary();
+    }
+
+    private function toUuid(string $binary): Uuid
+    {
+        return Uuid::fromString($binary);
+    }
```
chr-hertel, line 152: what's the purpose of those methods? we could just inline that, can't we?

chr-hertel (approved):
Awesome! Thanks for working on this.
Just one minor question - and what do you think about adding a `compose.yaml` to bring MariaDB to the repo for dev setup?

src/Store/Bridge/MariaDB/Store.php
valtzu: Sure – I used MongoDB store as a template and it had these

I'll inline

src/Store/Bridge/MariaDB/Store.php
```diff
@@ -0,0 +1,153 @@
...
+{
+    /**
+     * @param string $databaseName    The name of the database
+     * @param string $tableName       The name of the table
+     * @param string $indexName       The name of the vector search index
+     * @param string $vectorFieldName The name of the field int the index that contains the vector
+     */
+    public function __construct(
+        private \PDO $connection,
+        private string $databaseName,
```
valtzu, line 32: @chr-hertel what do you think about `$databaseName` – is it necessary? I think most of the time a connection has pre-selected database and passing this from Doctrine DSN etc could be difficult

I'm considering dropping this

composer.json
```diff
@@ -37,9 +37,11 @@
         "webmozart/assert": "^1.11"
     },
     "require-dev": {
+        "ext-pdo": "*",
         "codewithkyrian/chromadb-php": "^0.2.1 || ^0.3 || ^0.4",
         "codewithkyrian/transformers": "^0.5.3",
         "async-aws/bedrock-runtime": "^0.1.0",
+        "doctrine/dbal": "^4.2",
```
chr-hertel, line 44: could also be, right?
```suggestion
        "doctrine/dbal": "^3.0 || ^4.0",
```

we won't see a difference tho without example or test :D

PR #347, opened 2025-06-26
feat: Add Gemini Embeddings

valtzu:
Add Embeddings support for Google Gemini.

Only batch embedding implemented.

Related to #28

### DIFF
--- a/src/Platform/Bridge/Google/Embeddings.php
+++ b/src/Platform/Bridge/Google/Embeddings.php
@@ -1,0 +1,30 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpLlm\LlmChain\Platform\Bridge\Google;
+
+use PhpLlm\LlmChain\Platform\Bridge\Google\Embeddings\TaskType;
+use PhpLlm\LlmChain\Platform\Capability;
+use PhpLlm\LlmChain\Platform\Model;
+
+/**
+ * @author Valtteri R <valtzu@gmail.com>
+ */
+class Embeddings extends Model
+{
+    /** Supported dimensions: 3072, 1536, or 768 */
+    public const GEMINI_EMBEDDING_EXP_03_07 = 'gemini-embedding-exp-03-07';
+    /** Fixed 768 dimensions */
+    public const TEXT_EMBEDDING_004 = 'text-embedding-004';
+    /** Fixed 768 dimensions */
+    public const EMBEDDING_001 = 'embedding-001';
+
+    /**
+     * @param array{dimensions?: int, task_type?: TaskType|string} $options
+     */
+    public function __construct(string $name = self::GEMINI_EMBEDDING_EXP_03_07, array $options = [])
+    {
+        parent::__construct($name, [Capability::INPUT_MULTIPLE], $options);
+    }
+}
--- a/tests/Platform/Bridge/Google/Embeddings/EmbeddingsModelClientTest.php
+++ b/tests/Platform/Bridge/Google/Embeddings/EmbeddingsModelClientTest.php
@@ -1,0 +1,102 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpLlm\LlmChain\Tests\Platform\Bridge\Google\Embeddings;
+
+use PhpLlm\LlmChain\Platform\Bridge\Google\Embeddings;
+use PhpLlm\LlmChain\Platform\Bridge\Google\Embeddings\ModelClient;
+use PhpLlm\LlmChain\Platform\Response\VectorResponse;
+use PhpLlm\LlmChain\Platform\Vector\Vector;
+use PHPUnit\Framework\Attributes\CoversClass;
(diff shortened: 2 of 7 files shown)


tests/Platform/Bridge/Google/Embeddings/EmbeddingsModelClientTest.php
```diff
@@ -0,0 +1,77 @@
...
+{
+    #[Test]
+    public function itRequests(): void
+    {
+        $response = $this->createStub(ResponseInterface::class);
+        $response
+            ->method('toArray')
+            ->willReturn(json_decode($this->getEmbeddingStub(), true));
+
+        $httpClient = self::createMock(HttpClientInterface::class);
```
OskarStark, line 34: Also stub? We don't expect anything on it

valtzu: Now we actually expect something – it was quite dummy test before

valtzu:
Moved extra parameters to model's `$options` since those are mostly model-specific.

---

Also noticed that we probably need to parameterize `Store`s and pass `$dimensions` there, because trying to use those fixed-768-dimension models to 1536-dimension vector in MariaDB store fails with :point_down: 

```
Incorrect vector value: 'I\x16\xD7\xBC\x0F\xDB,\xBC\xFF\x0F\xA5<\x8C\xFE\x95\xBD~#^\xBC;\x84'\xBC\x11\xFD\x00<[\x82\x15<\xD5\x9A\x0A="\x12R;\x0A\x10~<...' for column `my_database`.`my_table`.`embedding`
```

But that'll of course be a separate PR.

examples/store/mariadb-similarity-search.php
OskarStark, line 1: What about keeping this example and add a separate one ?

src/Platform/Bridge/Google/Embeddings.php
```diff
@@ -0,0 +1,34 @@
...
+    /** Fixed 768 dimensions */
+    public const EMBEDDING_001 = 'embedding-001';
+
+    /**
+     * @param array{dimensions?: int, task_type?: TaskType|string, title?: string} $options
+     */
+    public function __construct(string $name = self::GEMINI_EMBEDDING_EXP_03_07, array $options = [])
+    {
+        if (self::GEMINI_EMBEDDING_EXP_03_07 === $name) {
+            $options['dimensions'] ??= 1536;
```
OskarStark, line 29: I am not a big fan of rebuilding the api validation, if it changes we need to keep up.

What happens if you send a wrong dimensions value?

tests/Platform/Bridge/Google/Embeddings/EmbeddingsModelClientTest.php
```diff
@@ -0,0 +1,102 @@
...
+use PHPUnit\Framework\TestCase;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+use Symfony\Contracts\HttpClient\ResponseInterface;
+
+#[CoversClass(ModelClient::class)]
+#[Small]
+#[UsesClass(Vector::class)]
+#[UsesClass(VectorResponse::class)]
+#[UsesClass(Embeddings::class)]
+class EmbeddingsModelClientTest extends TestCase
```
OskarStark, line 24:
```suggestion
final class EmbeddingsModelClientTest extends TestCase
```

src/Platform/Bridge/Google/Embeddings.php
valtzu: This is not really about api validation, but about the fact that the Store implementations (at least MariaDB) have hardcoded dimensions currently and thus do not work with default settings.

This can probably be removed after MariaDB store will have `dimensions` in `$options` of `initialize()`

OskarStark: Let's merge the options PR first then 👍🏻

valtzu: :point_up:  https://github.com/php-llm/llm-chain/pull/348

> I am not a big fan of rebuilding the api validation

Should I remove `TaskType` enum then also?

OskarStark: To me the enum is ok, WDYT @chr-hertel ?

chr-hertel: Yea, i also Like the idea of providing that - nice for DX.
We only should take into account not to block users if there are new task types released by Google.
But the implementation still support plain string for that option instead an enum?
With huggingface i went with an interface:
https://github.com/php-llm/llm-chain/blob/main/src/Platform/Bridge/HuggingFace/Task.php instead

valtzu: Ok, thanks – I kept the enum around – and yes, `string` is accepted too so we're not restricting future model/parameter usage.

Also rebased after #348, I think this PR is good to go now :rocket:

PR #348, opened 2025-06-26
chore: Allow arbitrary vector dimensions for MariaDB store

valtzu:
MariaDB insists the dimensions of the vector being inserted match exactly the dimensions of the vector defined in the schema.

Since the store knows nothing about the Model & it's expected vector dimensions, let's allow passing vector dimensions as a parameter to `initialize()` leave the problem to the user.


chr-hertel (approved):
LGTM, thank you @valtzu 👍

PR #351, opened 2025-06-27
fix: recursively remove `additionalProperties` from Gemini requests

valtzu:
Tool parameters with nested objects cause 400 bad request from Gemini due to only removing the unsupported `additionalProperties` on top-level.

```
Invalid JSON payload received. Unknown name "additionalProperties" at 'tools[0].function_declarations[0].parameters.properties[0].value': Cannot find field.
```


chr-hertel (approved):
Nice catch, thanks :+1:

PR #354, opened 2025-06-28
feat: Google structured output

valtzu:


chr-hertel (approved):
You are on fire - love it - thanks!! :+1:

PR #355, opened 2025-06-28
feat: Allow any normalizable result from a tool

### LINKED ISSUE #132: Empower tools to return whatever they want instead of `string`

valtzu:
Resolves #132

### DIFF
--- a/src/Chain/Toolbox/ToolResultConverter.php
+++ b/src/Chain/Toolbox/ToolResultConverter.php
@@ -4,32 +4,33 @@ declare(strict_types=1);
 
 namespace PhpLlm\LlmChain\Chain\Toolbox;
 
+use Symfony\Component\Serializer\Encoder\JsonEncoder;
+use Symfony\Component\Serializer\Normalizer\DateTimeNormalizer;
+use Symfony\Component\Serializer\Normalizer\JsonSerializableNormalizer;
+use Symfony\Component\Serializer\Normalizer\ObjectNormalizer;
+use Symfony\Component\Serializer\Serializer;
+use Symfony\Component\Serializer\SerializerInterface;
+
 /**
  * @author Christopher Hertel <mail@christopher-hertel.de>
  */
 final readonly class ToolResultConverter
 {
-    /**
-     * @param \JsonSerializable|\Stringable|array<int|string, mixed>|float|string|null $result
-     */
-    public function convert(\JsonSerializable|\Stringable|array|float|string|\DateTimeInterface|null $result): ?string
+    public function __construct(
+        private SerializerInterface $serializer = new Serializer([new JsonSerializableNormalizer(), new DateTimeNormalizer(), new ObjectNormalizer()], [new JsonEncoder()]),
+    ) {
+    }
+
+    public function convert(mixed $result): ?string
     {
-        if (null === $result) {
-            return null;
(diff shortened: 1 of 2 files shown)


src/Chain/Toolbox/ToolResultConverter.php
```diff
@@ -4,32 +4,43 @@
...
  * @author Christopher Hertel <mail@christopher-hertel.de>
  */
 final readonly class ToolResultConverter
 {
-    /**
-     * @param \JsonSerializable|\Stringable|array<int|string, mixed>|float|string|null $result
-     */
-    public function convert(\JsonSerializable|\Stringable|array|float|string|\DateTimeInterface|null $result): ?string
+    public function __construct(
+        private NormalizerInterface&EncoderInterface $serializer = new Serializer([new DateTimeNormalizer(), new ObjectNormalizer()], [new JsonEncoder()]),
```
chr-hertel, line 20: why not simply `SerializerInterface`?

src/Chain/Toolbox/ToolResultConverter.php
```diff
@@ -4,32 +4,43 @@
...
-        if ($result instanceof \DateTimeInterface) {
-            return $result->format(\DATE_ATOM);
+        $normalized = $this->serializer->normalize($result, 'json');
+
+        if (\is_scalar($normalized)) {
+            return (string) $normalized;
         }
 
-        return $result;
+        return $this->serializer->encode($normalized, 'json');
```
chr-hertel, line 44: and if you just replace those lines with

```php
return $this->serializer->serialize($result, 'json');
```

valtzu: F.e. date objects serialize to `string`, and and json it would be `"2025-06-28T23:09:20+00:00"` with quotes

This would mean different behavior than before

OskarStark: Can you add a testcase for this special case?

chr-hertel: i think that's quite fair - or do you see a downside in that?

makes it easier here, and if some want more control, they can just return a formatted string directly

...

as future iteration we could also think of exposing the `$context` array of the `serialize(...)` call via the tool's config maybe :thinking: - but that's rather a follow up i guess

valtzu: > do you see a downside in that?

Well, to me it feels a bit counter-intuitive that a `string` is rendered differently depending on if it was a `string` before or after normalization. It's not only about quotes but all other json-encoding, backslashes, unicode etc.

I'm quite new to this stuff so I can't say for sure if this would have any effect on the functionality. I agree it's easier/cleaner for this project if we can just use `serialize` as-is – so let's go with that :+1:

chr-hertel (approved):
Thank you @valtzu!

PR #356, opened 2025-06-28
feat: Google Gemini server tools

valtzu:
Add support for Google server-side tools like [Google search](https://ai.google.dev/gemini-api/docs/google-search), [URL context](https://ai.google.dev/gemini-api/docs/url-context) & [Code execution](https://ai.google.dev/gemini-api/docs/code-execution)

---

I thought of using `Tool` classes for this but didn't find any solution I was happy with, so went the easy way for now.

Ideally no changes would've been required for this approach at all, but there were 2 issues:
1. Gemini API insist on providing empty object, and IMO it's not feasible to expect end-user to provide `new \ArrayObject()`
2. If you provide `tools` in `call`'s `$options`, then all toolbox tools go away – hence the new `server_tools` _(open to better naming – `remote_tools`?)_

---

If someone has already given this a thought, please share your thoughts / if there is some clear better approach

### DIFF
--- a/examples/google/server-tools.php
+++ b/examples/google/server-tools.php
@@ -1,0 +1,40 @@
+<?php
+
+use PhpLlm\LlmChain\Chain\Chain;
+use PhpLlm\LlmChain\Chain\Toolbox\ChainProcessor;
+use PhpLlm\LlmChain\Chain\Toolbox\Tool\Clock;
+use PhpLlm\LlmChain\Chain\Toolbox\Toolbox;
+use PhpLlm\LlmChain\Platform\Bridge\Google\Gemini;
+use PhpLlm\LlmChain\Platform\Bridge\Google\PlatformFactory;
+use PhpLlm\LlmChain\Platform\Message\Message;
+use PhpLlm\LlmChain\Platform\Message\MessageBag;
+use Symfony\Component\Dotenv\Dotenv;
+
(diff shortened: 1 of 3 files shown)


chr-hertel:
Yes, another good one :) came across that with Claude as well - also called "server tools" there - so i guess the naming is good :+1: 
https://docs.anthropic.com/en/docs/agents-and-tools/tool-use/overview#server-tools

examples/google/server-tools.php
```diff
@@ -0,0 +1,39 @@
...
+$messages = new MessageBag(
+    Message::ofUser(
+        <<<'PROMPT'
+            What was the 12 month Euribor rate a week ago based on https://www.euribor-rates.eu/en/current-euribor-rates/4/euribor-rate-12-months/
+            PROMPT,
+    ),
+);
+
+// Available server-side tools as of 2025-06-28: url_context, google_search, code_execution
+$response = $chain->call($messages, ['server_tools' => ['url_context' => true]]);
```
chr-hertel, line 37: what do you think about enabling those server tools on the model class instead? feels a bit more like a feature of that model than of the chain to me

chr-hertel:
tested that all other toolcall examples still work with the new serializer setting - but with that server-tool example i always get a 429 response - guess that's due to my google setup atm

chr-hertel (approved):
Thanks @valtzu :)

PR #359, opened 2025-06-29
feat: Denormalize tool arguments

### LINKED ISSUE #352: Tool call parameter denormalization
While it's possible to create a tool that takes an object as a parameter, and the json schema gets created correctly, when doing the tool call, it fails with

```
Uncaught TypeError: class@anonymous(): Argument #1 ($dto) must be of type Dto, array given, called in /src/llm-chain/src/Chain/Toolbox/Toolbox.php on line 73 
```

Would it be possible to denormalize the parameters? Probably inside `ToolBox\ChainProcessor::processInput()`?

---

Not sure if this is a bug or a feature request – or maybe I just didn't find the correct pieces to wire together.

valtzu:
Resolves #352

### DIFF
--- a/src/Chain/Toolbox/Event/ToolCallArgumentsResolved.php
+++ b/src/Chain/Toolbox/Event/ToolCallArgumentsResolved.php
@@ -1,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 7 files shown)


chr-hertel (approved):
Thanks @valtzu - great and merged :)

PR #361, opened 2025-06-29
Add `CLAUDE.md` file

OskarStark:
I added the `.claude` directory to my global `.gitignore`


chr-hertel:
Had the same already open with #337, but decided against since this is a temporary resource depending on the repository code - basically like a build that would need update.
but only people with claude code could update that - and interestingly enough the delta is quite huge between our versions:
https://www.diffchecker.com/l7N4brNW/

OskarStark:
fine then

PR #364, opened 2025-06-29
feat\!: add unique ID to all messages

### LINKED ISSUE #77: Introduce UID for Messages

OskarStark:
## Summary
- Added unique identifiers (UUIDv7) to all message types
- Each message now automatically gets a unique ID upon instantiation
- Added comprehensive tests for the new ID functionality

## Breaking Change 🚨
This is a breaking change as `MessageInterface` now requires the `getId(): Uuid` method to be implemented by all message classes.

## Implementation Details
- Added `symfony/uid` package dependency (^7.3)
- Added `public readonly Uuid $id` property to all message classes
- IDs are generated automatically in constructors using `Uuid::v7()`
- Added `getId()` method to all message implementations

## Test Coverage
Added tests for each message type to ensure:
- ID is properly generated and accessible
- ID remains consistent for the same message instance
- Different message instances have different IDs

Closes #77
Closes #344

🤖 Generated with [Claude Code](https://claude.ai/code)


OskarStark:
## Why UUID v7?

UUID v7 provides key advantages for message identification:

- **Time-ordered**: Built-in timestamp allows natural chronological sorting without additional fields
- **Millisecond precision**: Captures creation time with high accuracy
- **Collision-resistant**: Combines timestamp with randomness for uniqueness even at high generation rates
- **Database-friendly**: Sequential nature improves index performance compared to random UUIDs

README.md
```diff
@@ -89,6 +89,31 @@ a **MessageBag** to a **Chain**, which takes care of LLM invocation and response
 Messages can be of different types, most importantly `UserMessage`, `SystemMessage`, or `AssistantMessage`, and can also
 have different content types, like `Text`, `Image` or `Audio`.
 
+#### Message Unique IDs
+
+Each message automatically receives a unique identifier (UUID v7) upon creation. This provides several benefits:
```
OskarStark, line 94:
```suggestion
Each message automatically receives a unique identifier (UUID v7) upon creation.
This provides several benefits:
```

OskarStark:
Hi @OskarStark,

In our project, we also need the 'timestamp' key in the serialized format. I'm unsure if this should be part of the library or if we should decorate the normalizers in our project.

Since UUID v7 already contains timestamp information that can be extracted, should we include a 'timestamp' field in the normalized message format? This would make the timestamp easily accessible without requiring users to parse the UUID.

What are your thoughts on this?

OskarStark:
Great discussion\! I think Dennis's suggestion for configurable timestamps is excellent and would provide the flexibility needed.

However, I believe we should proceed step by step. This PR already implements the core UUID functionality and ID serialization, which is the foundation. Let's get this merged first, then we can tackle the configurable timestamp feature in a follow-up PR.

The current implementation gives us:
- Unique message IDs (UUID v7 with embedded timestamps)
- ID serialization in all normalizers
- Solid foundation for future timestamp features

Once this is merged, we can implement the configurable timestamp context approach in a separate, focused PR. What do you think?

tests/Platform/ContractTest.php
```diff
@@ -61,7 +64,42 @@ public function createRequestPayload(Model $model, array|string|object $input, a
...
-        self::assertSame($expected, $actual);
+        // For array comparisons, we need to handle the dynamic ID fields
+        if (\is_array($expected) && \is_array($actual)) {
+            $this->assertPayloadEquals($expected, $actual);
+        } else {
+            self::assertSame($expected, $actual);
+        }
+    }
+
+    private function assertPayloadEquals(array $expected, array $actual): void
```
OskarStark, line 75: I don't like this TBH 😄

OskarStark:
CI 💚

OskarStark:
So I could get an approval from you @DZunke ? 😄

DZunke (approved):

DZunke:
@OskarStark sure, i scrolled through the code a second time and yeah, it looks fine from my side.

chr-hertel:
So, @OskarStark, how much of code and comments were hand crafted and what AI? :laughing:

chr-hertel (changes requested):
The normalizers are intended for the contract of the API payload - so we cannot just add the ID there. OpenAI might tolerate that, but for example Mistral and Anthropic fail with invalid request errors.
The UID is intended for user land code anyways, and i guess after a revert in normalizers this is good to be merged :)

PR #368, opened 2025-06-30
fix: resolve array tool arguments

chr-hertel:
@valtzu any idea about this case?

found it after merging #359 with example [examples/toolbox/tavily.php](https://github.com/php-llm/llm-chain/blob/main/examples/toolbox/tavily.php)

patch feels a bit hacky, but does the job atm


valtzu:
Thanks! I think there is some room to improve overall this denormalization, to support object arrays and other things that Symfony normally supports. Probably would need to use typeinfo instead of raw reflection

Issue #371, opened 2025-07-01
Enhancing Message Extensibility and Serializer Customization


llupa:
**Context:**
The recent integration of the Symfony Serializer has standardized object normalization within the library. However, the current design presents challenges for developers aiming to extend message functionalities with business-specific metadata, such as timestamps for session continuity.

**Identified Challenges:**

1. **Final Message Class:**
   The `*Message` class is declared as `final`, preventing inheritance and thus limiting the ability to introduce custom properties or behaviors tailored to specific business needs.

```php
final readonly class SystemMessage implements MessageInterface
```
Similar final declarations exist for `UserMessage`, `AssistantMessage`, and `ToolCallMessage`. Because these classes cannot be extended, it is difficult to attach additional state (timestamp example) while still leveraging the library’s serializer.

2. **Serializer Accessibility:**
   The serializer instance is currently private and inlined within the `Platform` class, restricting the injection of custom normalizers or encoders necessary for handling extended message structures.

```php
public function __construct(
    iterable $modelClients,
    iterable $responseConverter,
    private ?Contract $contract = null,
) {
    $this->contract = $contract ?? Contract::create();
```

```php
public static function create(NormalizerInterface ...$normalizer): self
{
    // Messages
    $normalizer[] = new MessageBagNormalizer();
    $normalizer[] = new AssistantMessageNormalizer();
    ...
    return new self(
        new Serializer($normalizer),
    );
}

```

3. **Rigid Normalizer Configuration:**

    Since there is no public API to inject additional normalizers after a Platform is constructed, customizing serialization for extended message classes requires rebuilding the entire serializer.

   Adding custom normalizers requires modifications across multiple platform factories or constructors, increasing the complexity and maintenance overhead for developers seeking to customize serialization behavior.

```php
PhpLlm\LlmChain\Platform\Platform.php
PhpLlm\LlmChain\Platform\Bridge\Anthropic\PlatformFactory.php
PhpLlm\LlmChain\Platform\Bridge\Azure\OpenAI\PlatformFactory.php
PhpLlm\LlmChain\Platform\Bridge\Bedrock\Platform.php // sidenote: no factory?
PhpLlm\LlmChain\Platform\Bridge\Google\PlatformFactory.php
PhpLlm\LlmChain\Platform\Bridge\HuggingFace\PlatformFactory.php
PhpLlm\LlmChain\Platform\Bridge\Mistral\PlatformFactory.php
[...]

PR #372, opened 2025-07-01
feat: open contract for more flexible normalizer configuration

DZunke:
Could fix part of #371 - the question for flexibility.

Why? Messages are given on platform calls and are given with the `MessageInterface`. So it is very flexible what could be given on call. Normalizers are only configured once, internally. So one side is highly flexible, the other side not. 

Giving the contract an interface allows creating own contracts with whatever backed technology one wants. So more flexible. Additionally it allows to add a contract that has an amount of normalizers one wants. For making it easier i thought about having model contract sets. I did it for anthropic, google and the platform itself. 

So a custom contract could look like

```php

use PhpLlm\LlmChain\Platform\Bridge\OpenAI\PlatformFactory;
use PhpLlm\LlmChain\Platform\Contract;

$platform = PlatformFactory::create(
    'my-api-key',
    contract: Contract::create(new MyOwnMessageBagNormalizer(), new MyOwnMessageThingyNormalizer()),
);
```

or, totally flexible:

```php
use PhpLlm\LlmChain\Platform\Bridge\OpenAI\PlatformFactory;

$platform = PlatformFactory::create(
    'my-api-key',
    contract: new MyVeryOwnContractUtilizingJsonStreamer(),
);
```

Maybe this could be a way for more flexibility with the combination of non configurable normalizer and "unknown" message input?


src/Platform/Bridge/Anthropic/PlatformFactory.php
```diff
@@ -4,15 +4,9 @@
...
 
-use PhpLlm\LlmChain\Platform\Bridge\Anthropic\Contract\AssistantMessageNormalizer;
-use PhpLlm\LlmChain\Platform\Bridge\Anthropic\Contract\DocumentNormalizer;
-use PhpLlm\LlmChain\Platform\Bridge\Anthropic\Contract\DocumentUrlNormalizer;
-use PhpLlm\LlmChain\Platform\Bridge\Anthropic\Contract\ImageNormalizer;
-use PhpLlm\LlmChain\Platform\Bridge\Anthropic\Contract\ImageUrlNormalizer;
-use PhpLlm\LlmChain\Platform\Bridge\Anthropic\Contract\MessageBagNormalizer;
-use PhpLlm\LlmChain\Platform\Bridge\Anthropic\Contract\ToolCallMessageNormalizer;
-use PhpLlm\LlmChain\Platform\Bridge\Anthropic\Contract\ToolNormalizer;
+use PhpLlm\LlmChain\Platform\Bridge\Anthropic\Contract\AnthropicSet as AnthropicContractSet;
```
OskarStark, line 7: do we need the alias?

src/Platform/Bridge/Anthropic/Contract/AnthropicSet.php
```diff
@@ -0,0 +1,28 @@
...
+declare(strict_types=1);
+
+namespace PhpLlm\LlmChain\Platform\Bridge\Anthropic\Contract;
+
+use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
+
+/**
+ * @author Denis Zunke <denis.zunke@gmail.com>
+ */
+final readonly class AnthropicSet
```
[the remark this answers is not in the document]
DZunke: No. I have seen that there are Contract Builds that combine mixed model contract normalizers so my thought here was to give them as a "set" like rector is doing it for some tools. So you could combine the set. Maybe this is too much? 🤔

src/Platform/Bridge/Anthropic/Contract/AnthropicSet.php
```diff
@@ -0,0 +1,28 @@
...
+            new DocumentNormalizer(),
+            new DocumentUrlNormalizer(),
+            new ImageNormalizer(),
+            new ImageUrlNormalizer(),
+            new MessageBagNormalizer(),
+            new ToolCallMessageNormalizer(),
+            new ToolNormalizer(),
+        ];
+    }
+}
```
chr-hertel, line 28: how about just 

```php
final class AnthropicContract extends Contract
{
    public static function create(): self
    {
        return parent::create([
            new AssistantMessageNormalizer(),
            new DocumentNormalizer(),
            new DocumentUrlNormalizer(),
            new ImageNormalizer(),
            new ImageUrlNormalizer(),
            new MessageBagNormalizer(),
            new ToolCallMessageNormalizer(),
            new ToolNormalizer(),
        ]);
    }
}
```

would need some changes with the parent, but would ease the handling while instantiating the Platform:
```php
$contract ?? AnthropicContract::create()
```

src/Platform/Bridge/Google/Contract/GoogleSet.php
```diff
@@ -0,0 +1,25 @@
...
+declare(strict_types=1);
+
+namespace PhpLlm\LlmChain\Platform\Bridge\Google\Contract;
+
+use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
+
+/**
+ * @author Denis Zunke <denis.zunke@gmail.com>
+ */
+final readonly class GoogleSet
```
chr-hertel, line 12: maybe `GeminiSet`?

src/Platform/Bridge/Anthropic/Contract/AnthropicSet.php
DZunke: Thanks! This feels like a good improvement alternative to the rector inspired sets 😊

chr-hertel (approved):
Thanks @DZunke - code looks good & functional tests are good as well :)

PR #374, opened 2025-07-02
chore: remove occurrences of `empty()`

alexandre-daubois:
`empty()` is used less and less in favor of bool casts on arrays. Symfony, for example, banned it from its codebase. I suggest we remove them from here as well.


src/Platform/Bridge/Bedrock/Anthropic/ClaudeHandler.php
```diff
@@ -73,7 +73,7 @@ public function convert(InvokeModelResponse $bedrockResponse): LlmResponse
                 $toolCalls[] = new ToolCall($content['id'], $content['name'], $content['input']);
             }
         }
-        if (!empty($toolCalls)) {
+        if (!(!$toolCalls)) {
```
OskarStark, line 76:
```suggestion
        if (!$toolCalls) {
```
?

OskarStark (approved):

src/Platform/Bridge/Bedrock/Anthropic/ClaudeHandler.php
alexandre-daubois: Of course! I missed this one. These AI tools have still some way to go... :)

DZunke:
May i ask why those bool casts are favored instead of replacing `empty` by explicit checks? I understand the downsides of `empty` but it feels weared to check if an empty array is `false` instead of explicit check like `\count($foo) === 0`. 

I just try to understand the pro and con of this variant 🤔

alexandre-daubois:
I'm in favor of removing `empty()`, but I'm also fine using `\count($foo) === 0` to check array emptiness. The main difference is that by using a bool cast, this check is performed at the engine level instead of userland, which makes it more efficient. I can understand performance in not what matters the most here, and I can update the PR to use `count()` if you prefer. As long as `empty()` is avoided 😄

alexandre-daubois:
Alright! Explicit checks added on arrays

alexandre-daubois:
Ah yes of course, done 🙂

OskarStark (approved):

PR #376, opened 2025-07-04
refactor!: improve types in platform api

chr-hertel:
I was just tired of doing those `assert($response instanceof ...);` statements
also resolves the special status of `AsyncResponse` by decoupling it from the `ResponseInterface`

WDYT?


src/Platform/Response/ResponsePromise.php
```diff
@@ -0,0 +1,114 @@
...
+
+use PhpLlm\LlmChain\Platform\Exception\UnexpectedResponseTypeException;
+use PhpLlm\LlmChain\Platform\ResponseConverterInterface;
+use PhpLlm\LlmChain\Platform\Vector\Vector;
+use Symfony\Contracts\HttpClient\ResponseInterface as HttpResponse;
+
+/**
+ * @author Christopher Hertel <mail@christopher-hertel.de>
+ */
+final class ResponsePromise
```
smnandre, line 15: ❤

DZunke:
This looks pretty cool! Thanks @chr-hertel - those assertions you mentioned are all over the place and this looks like a cool solution. Mostly one knows what kind of request was sent and what should be the response but for type safety checks where required. Wrapping those checks to the library with the `ResponsePromise` feels like a good step forward!

README.md
```diff
@@ -680,7 +680,7 @@ use PhpLlm\LlmChain\Platform\Bridge\OpenAI\Embeddings;
 
 $embeddings = new Embeddings($platform, Embeddings::TEXT_3_SMALL);
 
-$vectors = $platform->request($embeddings, $textInput)->getContent();
+$vectors = $platform->request($embeddings, $textInput)->asVectors();
```
OskarStark, line 683: Off topic, but what about returning a vector collection here? Or overkill?

PR #377, opened 2025-07-05
feat: allow JsonSerializable objects in contract

DZunke:
As brought up in #371 it could be a possible extension point to the contract to reintroduce the support for `JsonSerializable` objects when there are no normalizers for an object. 

So it would be a possible alternative for custom `MessageBagInterface`, `MessageInterface`, `ContentInterface`, etc. by allowing them to implement the `JsonSerializable` interface.


OskarStark (approved):

OskarStark:
Release needed 😅

PR #381, opened 2025-07-07
fix: Gemini system prompt

valtzu:
I started to getting 400 Bad Request, not sure if it was due to changing a model or what. Changing `system_instructions` into an array like in the [docs](https://ai.google.dev/gemini-api/docs/text-generation#system-instructions) fixed it.

Tested at least with `gemini-1.5-pro` and `gemini-2.5-pro` and `gemini-2.0-flash`, works fine.


OskarStark:
Makes sense based on the key name "parts"

PR #382, opened 2025-07-07
feat: Add `#[IsGrantedTool]` for tool access control

### LINKED ISSUE #360: Tool access control
Many times a tool invoked by chain performs an action that can be performed by the user manually too. This type of actions usually have some access control that may depend on the given parameters: f.e. in Symfony http context by defining `#[IsGranted('post:delete', 'post')]` on the controller.

Do you think it would make sense to have something similar here too, or it's outside of the scope here? I was thinking one possible solution could be to allow exposing controllers as tools. Of course you can already add `#[AsTool]` to an existing controller, but since the tool call does not go through Symfony HttpKernel, f.e. the `IsGranted` attribute is never checked. It'd be awesome if a single `IsGranted` attribute could cover both, the HTTP endpoint and also LLM Chain tool call.

If you think this is outside the scope, then it'd be nice to dispatch some `ToolCallArgumentsEvent` (analogous to [`ControllerArgumentsEvent`](https://github.com/symfony/http-kernel/blob/7.4/Event/ControllerArgumentsEvent.php)) just before making the actual tool call. This way it's easy to hook into the event in user land, check the attributes and throw some `AccessDeniedException`.

---

Of course it's always possible to just call `$this->security->isGranted(...)` inside the tool, but I find declarative access control rules nicer.

valtzu:
Resolves #360

### DIFF
--- a/src/Chain/Toolbox/Security/Attribute/IsGrantedTool.php
+++ b/src/Chain/Toolbox/Security/Attribute/IsGrantedTool.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace PhpLlm\LlmChain\Chain\Toolbox\Security\Attribute;
+
+use PhpLlm\LlmChain\Platform\Tool\Tool;
+use Symfony\Component\ExpressionLanguage\Expression;
+
+/**
+ * Checks if user has permission to access to some tool resource using security roles and voters.
+ *
+ * @see https://symfony.com/doc/current/security.html#roles
+ *
+ * @author Valtteri R <valtzu@gmail.com>
+ */
+#[\Attribute(\Attribute::IS_REPEATABLE | \Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD)]
+final class IsGrantedTool
+{
+    /**
+     * @param string|Expression                                                             $attribute     The attribute that will be checked against a given authentication token and optional subject
+     * @param array<mixed>|string|Expression|\Closure(array<string,mixed>, Tool):mixed|null $subject       An optional subject - e.g. the current object being voted on
+     * @param string|null                                                                   $message       A custom message when access is not granted
+     * @param int|null                                                                      $exceptionCode If set, will add the exception code to thrown exception
+     */
+    public function __construct(
+        public string|Expression $attribute,
+        public array|string|Expression|\Closure|null $subject = null,
+        public ?string $message = null,
+        public ?int $exceptionCode = null,
+    ) {
+    }
+}
--- a/tests/Fixture/Tool/ToolWithIsGrantedOnMethod.php
+++ b/tests/Fixture/Tool/ToolWithIsGrantedOnMethod.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace PhpLlm\LlmChain\Tests\Fixture\Tool;
+
+use PhpLlm\LlmChain\Chain\Toolbox\Security\Attribute\IsGrantedTool;
+use Symfony\Component\ExpressionLanguage\Expression;
+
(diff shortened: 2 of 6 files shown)


tests/Fixture/Tool/ToolWithIsGrantedOnMethod.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace PhpLlm\LlmChain\Tests\Fixture\Tool;
+
+use PhpLlm\LlmChain\Chain\Toolbox\Security\Attribute\IsGranted;
+use Symfony\Component\ExpressionLanguage\Expression;
+
+final class ToolWithIsGrantedOnMethod
```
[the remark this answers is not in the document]
valtzu: You mean different namespace?

OskarStark (approved):

src/Chain/Toolbox/Security/Attribute/IsGranted.php
chr-hertel, line 1: Can we use the Symfony one instead? it's basically the same code besides `Tool` as a subject - but I can't spot where you need that

valtzu: It would mean requiring `symfony/security-http` instead of just  `symfony/security-core`, and that pulls in a lot of dependencies, f.e. `symfony/http-foundation` and `symfony/http-kernel`.

Of course since it's an optional dependency, maybe that's ok? And in practice I think most people use this library with the bundle & likely have full Symfony setup anyway so :shrug: 

---

> it's basically the same code besides Tool as a subject

it's about what arguments the `Closure` gets. If we use the built-in `IsGranted`, then should we provide `Request` object to the callback like the [typehint](https://github.com/symfony/symfony/blob/7.4/src/Symfony/Component/Security/Http/Attribute/IsGranted.php#L29) says?

Also it could be misleading that an expression in a single `#[IsGranted]` attribute would get different parameters for evaluation depending on the context:
- symfony http request: `['request' => ...,' 'args' => ...]`
- php-llm tool call: `['tool' => ...,' 'args' => ...]`

---

Do you have any take on this @OskarStark ?

valtzu: I guess more analogous to `Request` would be a `ToolCall` instance, but for some reason I didn't include it in the `ToolCallArgumentsResolved` :thinking:

OskarStark:
If the class is not available we can throw an exception, no?

valtzu:
I tried adding the check directly to the top of the `IsGrantedTool` file, but then realized it does not get executed before `$attr->newInstance()` call in the listener.

Can't think of any place except the toolbox itself where you could check it, and if you do, then it kinda defeats the purpose of using events to decouple logic/domains

src/Chain/Toolbox/Security/Attribute/IsGrantedTool.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace PhpLlm\LlmChain\Chain\Toolbox\Security\Attribute;
+
+use PhpLlm\LlmChain\Platform\Tool\Tool;
+use Symfony\Component\ExpressionLanguage\Expression;
+
+/**
+ * Checks if user has permission to access to some tool resource using security roles and voters.
+ *
```
OskarStark, line 10:
```suggestion
    throw new \LogicException('In order to use #[IsGrantedTool] attribute, you must install an event dispatcher and make sure IsGrantedToolAttributeListener is attached to it. Try running "composer require symfony/event-dispatcher"');
```

valtzu:
> should go to the bundle

Sounds good to me

PR #386, opened 2025-07-10
feat: expose Gemini grounding metadata

valtzu:
Example:

```
$ php examples/google/grounding.php
Symfony is an open-source PHP framework designed to accelerate the creation and maintenance of web applications. It provides a set of reusable PHP components and bundles that simplify complex development tasks, allowing developers to build robust and scalable applications. By promoting best practices and design patterns, Symfony gives developers full control over the application's configuration and customization.
Search queries: what is Symfony framework
Sources: wikipedia.org, ntchosting.com, symfony.com, drupalize.me
```

Up for discussion if it'd be better to do some extra processing _(to provide a stable data structure for the user of this library)_ for the data or it's ok to expose as-is.

---

Also worth noting that I see this information as something you want to save with your chat (basically within `AssistantMessage`), so in that sense I'm not 100% sure if this is the correct place for this.

### DIFF
--- a/src/Platform/Bridge/Google/GroundingOutputProcessor.php
+++ b/src/Platform/Bridge/Google/GroundingOutputProcessor.php
@@ -1,0 +1,41 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpLlm\LlmChain\Platform\Bridge\Google;
+
+use PhpLlm\LlmChain\Chain\Output;
+use PhpLlm\LlmChain\Chain\OutputProcessorInterface;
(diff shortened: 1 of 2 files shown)


src/Platform/Bridge/Google/GroundingOutputProcessor.php
chr-hertel, line 1: That's quite straight forward to test, please add one

chr-hertel:
@valtzu do you still want to work on this here or move it to `symfony/ai`?

PR #387, opened 2025-07-11
Add basic setup for memory injections to system prompt

DZunke:
This PR introduces a flexible memory system that allows the LLM to recall contextual information that are permantent for the conversation. In difference to tools it can be always utilized, when `use_memory` option is not disabled. It would be possible to fetch memory by a tool with a system instruction like `always call the tool foo_bar` but, for me, this feels like a bad design to always force the model to do a tool call without further need. 

Currently i have added just two memory providers to show what my idea is. I could also think about a write layer to fill a memory, together with a read layer, this could, for example, be working good with a graph database and tools for memory handling. But these are ideas for the future. 

So for this first throw i hope you get what i was thinking about to reach. I decided to inject the memory to the system prompt, when it is available, instead of adding a second system prompt to the message bag. But this was just a 50/50 thinking. I tried both seem to be working equally, at least for open ai. I am open to change it back again. 

What do you think?

### DIFF
--- a/src/Chain/Memory/MemoryInputProcessor.php
+++ b/src/Chain/Memory/MemoryInputProcessor.php
@@ -45,13 +45,17 @@ final readonly class MemoryInputProcessor implements InputProcessorInterface
 
         $memory = '';
         foreach ($this->memoryProviders as $provider) {
-            $memoryMessage = $provider->loadMemory($input);
+            $memoryMessages = $provider->loadMemory($input);
 
-            if (null === $memoryMessage) {
+            if (0 === \count($memoryMessages)) {
                 continue;
             }
 
-            $memory .= \PHP_EOL.\PHP_EOL.$memoryMessage->content;
+            $memory .= \PHP_EOL.\PHP_EOL;
+            $memory .= implode(
+                \PHP_EOL,
+                array_map(static fn (Memory $memory): string => $memory->content, $memoryMessages),
+            );
         }
 
         if ('' === $memory) {
--- a/src/Chain/Memory/MemoryProviderInterface.php
+++ b/src/Chain/Memory/MemoryProviderInterface.php
@@ -11,5 +11,8 @@ use PhpLlm\LlmChain\Chain\Input;
  */
 interface MemoryProviderInterface
 {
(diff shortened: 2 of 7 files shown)


src/Chain/Memory/MemoryInputProcessor.php
```diff
@@ -0,0 +1,72 @@
...
+use PhpLlm\LlmChain\Platform\Message\Message;
+
+/**
+ * @author Denis Zunke <denis.zunke@gmail.com>
+ */
+final readonly class MemoryInputProcessor implements InputProcessorInterface
+{
+    private const MEMORY_PROMPT_MESSAGE = <<<MARKDOWN
+        # Conversation Memory
+        This is the memory i have found for this conversation. The memory has more weight to answer user input,
```
chr-hertel, line 18:
```suggestion
        This is the memory I have found for this conversation. The memory has more weight to answer user input,
```

src/Chain/Memory/MemoryProviderInterface.php
```diff
@@ -0,0 +1,15 @@
...
+namespace PhpLlm\LlmChain\Chain\Memory;
+
+use PhpLlm\LlmChain\Chain\Input;
+
+/**
+ * @author Denis Zunke <denis.zunke@gmail.com>
+ */
+interface MemoryProviderInterface
+{
+    public function loadMemory(Input $input): ?Memory;
```
chr-hertel, line 14: what do you think about having `Memory[]` as return type here? I can imagine use-cases defining multiple ones

chr-hertel:
Love it - thanks @DZunke!

I think adding it to the existing system prompt is the right way to go - not all providers support having multiple system prompts.

Can see this in action already with the bundle and using a custom MemoryProviderInterface implementation to inject user specific data :+1:

src/Chain/Memory/MemoryProviderInterface.php
DZunke: Mhm. My intention was that the memory providers provide a fully formatted memory string but you may be right that it could also deliver multiple fully formatted memory strings. At least when it is a usecase someone has in mind the possibility is already there and so i will extend it :)

